PR #39, opened 2019-06-20
Add links to Data Hub company contacts

paulgain:
* New Order - (OMIS)
* Investment project - (FDI)
* Investment project - (Non-FDI)
* Investment project - (CTI)

https://trello.com/c/L1FIfGce/1308-add-contact-links-into-investment-projects-omis-orders


paulgain:
> 👍 looks good. Did the snapshots not need updating?
Nope, I've just run them again and nothing was generated.

paulgain:
> In addition, I have reservations about the URL coming back from the activity stream because it takes the user outside the context of the company. ie `/contacts/id` rather than `/company/id/contacts/id`

Ali has said he has spoke to you and goes on further to say: "The plan is that this is something that will require some thought and we are going to fix it once we have put the requisite thought in it".

PR #40, opened 2019-06-21
Update service delivery card to be more specific

yeahfro:
## Change
Currently service deliveries are using "Completed interaction" but they should be using "Completed
service delivery"

Link text with `interaction` should use `service delivery` for service deliveries.

## Before
Screenshot 2019-06-21 at 13 25 33

## After
Screenshot 2019-06-21 at 13 25 16

### DIFF
--- a/src/activity-feed/activity-feed-cards/InteractionUtils.js
+++ b/src/activity-feed/activity-feed-cards/InteractionUtils.js
@@ -43,10 +43,12 @@ export default class CardUtils {
     const status = getStatus(activity)
     const badge = isServiceDelivery(activity) ? BADGE_LABELS.COMPLETED_SERVICE_DELIVERY : BADGE_LABELS[status.toUpperCase()]
     const isUpcoming = status === STATUS.UPCOMING
+    const typeText = isServiceDelivery(activity) ? 'service delivery' : 'interaction'
 
     return {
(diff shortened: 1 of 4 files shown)


src/activity-feed/activity-feed-cards/InteractionUtils.js
```diff
@@ -33,15 +33,23 @@ const getStatus = (activity) => {
   }
 }
 
+const setIsServiceDelivery = (activity) => {
```
paulgain: This isn't setting anything as it just returns a boolean - shouldn't it be called `isServiceDelivery`?

yeahfro: Yeah that's what I had but there were clashes with the variable name `isServiceDelivery`. Open to ideas

yeahfro: I could just change this line to:

```
    const isServiceDelivery = includes(get(activity, 'object.type'), 'dit:ServiceDelivery')
```

paulgain: Yeah, that would work albeit somewhat ugly. You called also call the function twice and do away with the `const` as the function is not computationally expensive.

PR #45, opened 2019-06-21
Add ESLint

rafenden:


src/activity-feed/activity-feed-cards/card/CardUtils.jsx
```diff
@@ -29,9 +33,9 @@ const createJobTitleMarkup = ({id, url, name, jobTitle}) => {
 }
 
 const getPeople = (activity, personSubType) => {
-  return map(filter(activity['object']['attributedTo'], ({type}) => {
+  return map(filter(activity.object.attributedTo, ({ type }) => {
```
paulgain: Consider using `  const { attributedTo } = activity.object` for consistency.

src/activity-feed/activity-feed-cards/InteractionUtils.js
```diff
@@ -41,7 +45,9 @@ const isServiceDelivery = (activity) => {
 export default class CardUtils {
   static transform(activity) {
     const status = getStatus(activity)
-    const badge = isServiceDelivery(activity) ? BADGE_LABELS.COMPLETED_SERVICE_DELIVERY : BADGE_LABELS[status.toUpperCase()]
+    const badge = isServiceDelivery(activity)
+      ? BADGE_LABELS.COMPLETED_SERVICE_DELIVERY
+      : BADGE_LABELS[status.toUpperCase()]
```
paulgain: If `status` is `null` this line will throw a `TypeError` when calling `toUpperCase()`.

src/activity-feed/activity-feed-cards/InteractionUtils.js
```diff
@@ -19,18 +19,22 @@ const BADGE_LABELS = {
...
+    const startTime = get(activity, 'object.startTime')
+    const isUpcoming = new Date(startTime) > new Date()
+    return isUpcoming ? STATUS.UPCOMING : STATUS.INCOMPLETE
   }
+
+  if (apiStatus === STATUS.COMPLETE) {
+    return STATUS.COMPLETE
+  }
+
+  return null
```
yeahfro: Do we need this `return null` here? The interaction should always be `draft` or `complete` so we get to this point, returning `undefined` is okay.

rafenden: We don't need to return `null`, it's a just a good practise and currently it's one of the ESLint rules

PR #51, opened 2019-06-26
Add semantic-relase

rafenden:


README.md
```diff
@@ -69,11 +76,47 @@ yarn test:update
...
-Caution: Remember to be on the latest master, yarn will increase the version number in `package.json` and will create a tag and a commit message including the new version.
+This project is using [Semantic Release](https://github.com/semantic-release/semantic-release).
 
-```bash
-yarn release
-git push
+When committing please always use the following pattern for you messages (scope and body are optional):
+
+```
+<type>(<scope>): <subject>
```
yeahfro: Do we need to follow this exact pattern? It is a move away from our current pattern. ie can `<type>` be added to the footer? And you say `scope and body are optional`, what about `<footer>` being optional?

rafenden: This is a widely used pattern by `semantic-release` projects, and yes, footer is also optional, I will update the comment.

yeahfro: Okay, this has come up in a discussion before as it essentially eats into the minimal amount of characters.

PR #67, opened 2019-07-02
Add "Show details for all activities" toggle

yeahfro:
## Change
Clicking the link will show details for all activities. The text will become "Hide details for all activities". Clicking the link again will return it to its original state with all details hidden.

### Before
Screenshot 2019-07-02 at 14 38 08

### After
Screenshot 2019-07-04 at 05 48 14

### DIFF
--- a/src/activity-feed/ActivityFeed.test.jsx
+++ b/src/activity-feed/ActivityFeed.test.jsx
@@ -3,6 +3,8 @@ import renderer from 'react-test-renderer'
 import MockDate from 'mockdate'
 import { uniqueId } from 'lodash'
 import { mount } from 'enzyme'
+import { Details } from 'govuk-react'
+
 import ActivityFeed from './ActivityFeed'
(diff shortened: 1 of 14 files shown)


src/activity-feed/ActivityFeed.test.jsx
```diff
@@ -128,4 +130,28 @@ describe('ActivityFeed', () => {
...
+      wrapper = mount((<ActivityFeed
+        totalActivities={1}
+        activities={[interactionActivityFixture]}
+      />))
+
+      wrapper.find('ShowDetails').at(0).simulate('click')
+    })
+
+    test('should change the link text', () => {
+      expect(wrapper.find('ShowDetails').at(0).prop('children')).toEqual([
```
paulgain: Consider setting `wrapper.find('ShowDetails').at(0)` so you don't have to search a second time.

yeahfro: @paulgain thank you for that suggestion. It's what I originally had actually and when you change the state of a component you actually need to find it again.

PR #70, opened 2019-07-03
Update font size to be consistent

yeahfro:
## Change
Updates the global font sizes to be consistent with GDS design system. The activity card should be consistent with existing collection cards.

## Before
Screenshot 2019-07-03 at 21 34 53

## After
Screenshot 2019-07-04 at 05 27 28

### DIFF
--- a/src/activity-feed/activity-feed-cards/card/CardTable.jsx
+++ b/src/activity-feed/activity-feed-cards/card/CardTable.jsx
@@ -12,7 +12,7 @@ const GovUkTable = styled(Table)`
     font-weight: normal;
     border: 0;
     padding: ${SPACING.SCALE_2};
-    font-size: 14px;
+    font-size: 16px;
   }
 `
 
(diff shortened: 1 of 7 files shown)


src/activity-feed/activity-feed-cards/card/CardTable.jsx
```diff
@@ -12,7 +12,7 @@ const GovUkTable = styled(Table)`
     font-weight: normal;
     border: 0;
     padding: ${SPACING.SCALE_2};
-    font-size: 14px;
+    font-size: 16px;
```
paulgain: Consider using `font-size: ${FONT_SIZE.SIZE_16};` for consistency.

yeahfro: Just saw this and it was a mistake. Now I have merged the PR in there will be a separate PR to fix some of the consistency/naming of styled components.

PR #73, opened 2019-07-05
Refactor Contacts and Advisers into a vertical list

paulgain:
Currently we only show a single Company contact or a single Adviser when
we should display a list of them.

Ensure the labelling is consistent across all activity cards for
Company contacts - i.e. Company contact(s)

**Before**

before

**After**

after

https://trello.com/c/jfrFN7X8/1301-refactor-contacts-and-advisers-into-a-comma-separated-list
https://trello.com/c/BVAeTADZ/1339-activity-feed-tweaks


src/activity-feed/activity-feed-cards/card/AdvisersList.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Link } from 'govuk-react'
+import styled from 'styled-components'
+
+const UList = styled('ul')`
```
yeahfro: We are moving to the same naming convention as `govuk-react` so in this instance it would be `StyledList`

src/activity-feed/activity-feed-cards/card/AdvisersList.jsx
```diff
@@ -0,0 +1,48 @@
...
+  margin: 0;
+  padding-left: 0
+`
+
+const ListItem = styled('li')`
```
yeahfro: In this instance it would be `StyledListItem`

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
```diff
@@ -6,42 +6,44 @@ import {
...
+      {name}, <Link href={`mailto:${emailAddress}`}>{formattedEmailAddress}</Link> {teamName}
     </span>
   )
 }
 
-const getPeople = (activity, personSubType) => {
-  return map(filter(activity.object.attributedTo, ({ type }) => {
-    return includes(type, `dit:${personSubType}`)
-  }), ({ id, url, name, 'dit:jobTitle': jobTitle, 'dit:emailAddress': emailAddress }) => {
+const getContacts = (activity) => {
```
yeahfro: An option here could have been to leave these functions consolidated and use `pickBy` on the returned `object`

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
```diff
@@ -6,42 +6,44 @@ import {
...
+const getAdvisers = (activity) => {
+  const { attributedTo } = activity.object
+  return map(filter(attributedTo, ({ type }) => {
+    return includes(type, 'dit:Adviser')
+  }), ({ id, name, 'dit:emailAddress': emailAddress, 'dit:team': team }) => {
+    return {
+      id,
+      name,
       emailAddress,
+      team: team ? team.name : null,
```
yeahfro: ie here you could just use `get` and then the `pick` would remove anything that is `undefined`

src/activity-feed/activity-feed-cards/Interaction.jsx
```diff
@@ -45,8 +50,8 @@ export default class Interaction extends React.PureComponent {
         >
           <CardTable rows={
             [
-              { header: 'Contact(s)', content: CardUtils.getPeopleAsList(activity, 'Contact') },
-              { header: 'Adviser(s)', content: CardUtils.getPeopleAsList(activity, 'Adviser') },
+              { header: 'Company contact(s)', content: <ContactsList contacts={contacts} /> },
```
rafenden: As discuss on earlier, I suggest using a single component here `<CardPeople>`

paulgain:
> Looks good. My main apprehension is the duplication with the rendering of the advisers/contacts.
> 
> As discussed I think it would be nicer to have a `PeopleList` component and then `AdviserListItem` and `ContactListItem` if they are to be differentiated.
> 
> Let me know what you think.

I've taken @rafenden through this problem and he agrees this is the optimal solution because Contacts and Advisers are very different. Lumping everything into the same class (i.e. PeopleList) and then all the logic into a function that has to deal with all the formatting is not ideal.

src/activity-feed/activity-feed-cards/card/AdvisersList.jsx
paulgain: 👍

src/activity-feed/activity-feed-cards/card/AdvisersList.jsx
paulgain: 👍

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
paulgain: Looking at this function you cannot differentiate between the types, you wouldn't know what the Adviser properties are, the same applies to the Contact properties. In order to ascertain what properties belong to what object you would need to look at the prop types definition on each list. Considering developers spend more time reading code than actually writing code I would suggest splitting them out. 

It's a trade off from making something more generic with less code (meaning there is less to maintain and test) against **READABILITY**. 

What's the point if it's generic but ambiguous?

```
const getPeople = (activity, personType) => {
  const { attributedTo } = activity.object
  return map(filter(attributedTo, ({ type }) => {
    return includes(type, `dit:${personType}`)
  }), ({ id, url, name, 'dit:jobTitle': jobTitle, 'dit:emailAddress': emailAddress, 'dit:team': team }) => {
    return pickBy({
      id,
      url,
      name,
      jobTitle,
      emailAddress,
      team: team ? team.name : null,
    })
  })
}
```

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
paulgain: 👍

src/activity-feed/activity-feed-cards/Interaction.jsx
paulgain: As discussed, demonstrated and agreed on yesterday it's far better to define two lists, one for Contacts, the other for Advisers.

src/activity-feed/activity-feed-cards/Omis.jsx
```diff
@@ -50,7 +52,7 @@ export default class Omis extends React.PureComponent {
               { header: 'Country', content: country },
               { header: 'UK region', content: ukRegion },
               { header: 'Added by', content: CardUtils.getAddedBy(activity) },
```
rafenden: Could we use here the `<Adviser />` component?

paulgain: Yes, that's the next piece of work, it's on a separate ticket.

PR #81, opened 2019-07-10
Refactor AddedBy (Adviser)

paulgain:
Refactored `AddedBY` (which is an Adviser) so it reuses the same elements defined in the Adviser list. 

**There are no visual changes.**


src/activity-feed/activity-feed-cards/InvestmentProject.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import { get } from 'lodash'
 import PropTypes from 'prop-types'
+import Adviser from './card/list/adviser/Adviser'
```
yeahfro: This `import` would need to be with our other `import` statements which are not third party/

src/activity-feed/activity-feed-cards/InvestmentProject.jsx
```diff
@@ -40,7 +41,8 @@ export default class InvestmentProject extends React.PureComponent {
     const url = get(activity, 'object.url')
     const name = get(activity, 'object.name')
     const investmentType = get(activity, 'object.dit:investmentType.name')
-    const addedBy = CardUtils.getAddedBy(activity)
+    const adviser = CardUtils.getAddedBy(activity)
+    const adviserComponent = adviser ? <Adviser adviser={adviser} /> : null
```
yeahfro: Could your `Adviser` component just return `null` based on the `adviser` parameter?

yeahfro: In addition, this would be named `AdviserComponent` as it is a component, however I feel it could be cleaner if `Adviser` returns `null`.

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
```diff
@@ -73,8 +56,15 @@ export default class CardUtils {
   }
 
   static getAddedBy = (activity) => {
```
yeahfro: Should this still be called `getAddedBy` or is it actually `getAdviser`? ie is the `actor` not necessarily the `adviser`?

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
```diff
@@ -73,8 +56,15 @@ export default class CardUtils {
...
-    const name = get(activity, 'actor.name')
-    const emailAddress = get(activity, 'actor.dit:emailAddress')
-    return createEmailAddressMarkup({ name, emailAddress })
+    const adviser = {
```
yeahfro: Based on previous comment if this is not `getAdviser` then is this the right name `adviser`?

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
```diff
@@ -73,8 +56,15 @@ export default class CardUtils {
...
+      name: get(activity, 'actor.name'),
+      emailAddress: get(activity, 'actor.dit:emailAddress'),
+    }
+
+    if (!adviser.name || !adviser.emailAddress) {
```
yeahfro: Would something like this be an option?

```
return adviser.name && adviser.emailAddress ? adviser : null
```

src/activity-feed/activity-feed-cards/card/list/adviser/Adviser.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { Link } from 'govuk-react'
+import PropTypes from 'prop-types'
+
+import PROPS from './props'
+
+const Adviser = ({ adviser }) => {
```
yeahfro: This `class` is using a slightly different pattern to other classes. Should we just make it consistent?

src/activity-feed/activity-feed-cards/card/list/adviser/AdvisersList.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
```
yeahfro: Aren't `AdvisersList` and `ContactsList` going to be the same as it's just the `*ListItem` that is changing?

src/activity-feed/activity-feed-cards/card/list/adviser/props.js
```diff
@@ -0,0 +1,9 @@
+import PropTypes from 'prop-types'
```
yeahfro: Again this pattern is slightly different. I quite like seeing the `PropTypes` in the `class` rather than digging around in an abstraction. Is there another benefit of doing this?

src/activity-feed/activity-feed-cards/InvestmentProject.jsx
paulgain: > Could your `Adviser` component just return `null` based on the `adviser` parameter?

In short, React doesn't work like that.

```
const Adviser = ({ adviser }) => {
  if (!adviser) {
    return null
  }
  ...
}
```

```
const adviser = CardUtils.getAddedBy(activity)
```

```
    return (
      <Card>
          ...
          <CardTable rows={
            [
              ...
              { header: 'Added by', content: <Adviser adviser={adviser} /> },
              ...
            ]}
          />
        </CardDetails>
      </Card>
    )
```
Unfortunately React doesn't do what you would expect here. If `adviser` is null React will create an element that has a prop type called adviser and set it to null. 

**JSX Example**
```
const Adviser = ({ adviser }) => {
  if (!adviser) {
    return null
  }

  return <div>Test</div>
}

const adviserComponent = <Adviser adviser={null} />
```

**JSX conversion via Babel**
```
const Adviser = ({
  adviser
}) => {
  if (!adviser) {
    return null;
  }

  return React.createElement("div", null, "Test");
};

const adviserComponent = React.createElement(Adviser, {
  adviser: null
});
```
The `CardTable` component renders the row with the heading `Added by` and also the empty react element so you end up with this:

empty

src/activity-feed/activity-feed-cards/card/CardUtils.jsx
paulgain: In this scenario the actor is always the Adviser. I'll rename the function, it's not a problem.

src/activity-feed/activity-feed-cards/card/list/adviser/Adviser.jsx
paulgain: Its not a class it's a React function: https://reactjs.org/docs/components-and-props.html

Making everything a class is not the correct thing to do. A class has state and additional Lifecycle hooks that are simply not necessary for this scenario (i.e. an Adviser).
https://reactjs.org/docs/state-and-lifecycle.html

The rule of thumb should be to use a class if your component has state, and or where you need to utilise the Lifecycle hooks or override them otherwise use a React function as they are much simpler and come with less overhead.

I've been guilty myself of using a class when a function is the better choice.

src/activity-feed/activity-feed-cards/card/list/adviser/AdvisersList.jsx
paulgain: Yes that is true. 

They are quite similar but they differ on the prop types so it depends on how strict we want to be with that. This is how the prop types between Contact and Adviser differ

**Contact**
```
id,
name,
url,
jobTitle, // Optional field
```

**Adviser**
```
id,
name,
emailAddress,
team: // Only available for Interactions
```

If we had a generic list how would the prop types work? I genuinely don't know. I like the idea of being strict with the prop types because you know by reading the code exactly what each object properties are.

src/activity-feed/activity-feed-cards/card/list/adviser/props.js
paulgain: It's reusability, why define the same prop types more than once? This is used three times so it makes sense to pull it out into its own file where it's defined just once. 

What are the benefits of duplicating code?

paulgain:
> I have left a few comments. I feel like a `CardDetailsList` component would work well as it's just the list item presentation that is changing. ie we don't want to define a new `List` component each time we want a list in the details table.

I have already gone through this with @rafenden please speak to him as I'm not wasting any more of my time on this.

src/activity-feed/activity-feed-cards/card/index.js
```diff
@@ -4,5 +4,5 @@ export { default as CardHeader } from './CardHeader'
 export { default as CardHeading } from './CardHeading'
 export { default as CardMeta } from './CardMeta'
 export { default as CardTable } from './CardTable'
-export { default as ContactsList } from './ContactsList'
-export { default as AdvisersList } from './AdvisersList'
+export { default as ContactsList } from './list/ContactsList'
```
rafenden: Do we need this additional nesting?

src/activity-feed/activity-feed-cards/card/list/adviser/Adviser.jsx
rafenden: > The rule of thumb should be to use a class if your component has state

Do you have a reference to that rule?

src/activity-feed/activity-feed-cards/card/list/adviser/props.js
rafenden: In this case I would say it's fine as we don't duplicate any logic, only schema.

src/activity-feed/activity-feed-cards/card/list/adviser/AdvisersList.jsx
yeahfro: Wouldn't the differentiated `PropTypes` just be on the list item? ie what's the point in [this](https://github.com/uktrade/data-hub-components/pull/81/files#diff-cdb853355915f04a53809d77c1074b60R16) when we have [this](https://github.com/uktrade/data-hub-components/pull/81/files#diff-3733ba03a58fca7eb1aa1bac85eb8fdcR14)

src/activity-feed/activity-feed-cards/InvestmentProject.jsx
yeahfro: Why can't we achieve the same as [this line](https://github.com/uktrade/data-hub-components/blob/e2fe59bbb24833b770419ca0e1f39fd8684ca440/src/activity-feed/ActivityFeedCard.jsx#L23)?

src/activity-feed/activity-feed-cards/card/list/adviser/Adviser.jsx
paulgain: Seriously, this is common sense. You are extending a class, therefore, you are inheriting state and behaviour and the overhead that comes with it - software engineering 101. I'm not happy that I have to find a reputable reference when as a software developer you should know this: 
https://twitter.com/dan_abramov/status/993103559297204224

src/activity-feed/activity-feed-cards/InvestmentProject.jsx
paulgain: 👍

src/activity-feed/activity-feed-cards/card/list/adviser/AdvisersList.jsx
paulgain: "Wouldn't the differentiated PropTypes just be on the list item?". Yes, that makes sense. What `PropTypes ` do we define at the generic list level?

_"ie what's the point in this when we have this"_

When building out lists React only requires ids at the highest level, in our case at the list level. https://reactjs.org/docs/lists-and-keys.html

```
function ListItem(props) {
  // Correct! There is no need to specify the key here:
  return <li>{props.value}</li>;
}

function NumberList(props) {
  const numbers = props.numbers;
  const listItems = numbers.map((number) =>
    // Correct! Key should be specified inside the array.
    <ListItem key={number.toString()}
              value={number} />

  );
  return (
    <ul>
      {listItems}
    </ul>
  );
}

const numbers = [1, 2, 3, 4, 5];
ReactDOM.render(
  <NumberList numbers={numbers} />,
  document.getElementById('root')
);
```

src/activity-feed/activity-feed-cards/card/index.js
paulgain: It's called project structure, it makes things easier to find rather than dumping everything in a single directory.

src/activity-feed/activity-feed-cards/card/list/adviser/Adviser.jsx
yeahfro: @paulgain it would be unfair to say that this is "software engineering 101" as that would imply this would be the rule for all object oriented programming languages which is not the case.

paulgain: You take my point, there are always exceptions to every rule. I prefer functions over classes within this **context** for exactly that reason. We don't need to role out classes everywhere for consistency. Can we please agree on that?

rafenden: > I'm not happy that I have to find a reputable reference when as a software developer you should know this

I asked about this because I couldn't find the reference myself. By reference a mean a documentation on the official docs.

rafenden:
Please use the commit message which complies with the semantic release format, e.g.
```
refactor: Refactor AddedBy (Adviser)
```

PR #84, opened 2019-07-16
refactor: refactor AddedBy (Adviser)

paulgain:
Refactored Added By so it reuses the same elements defined in the Adviser list.

Refactored both Contacts list and Advisers list into a single list called CardDetailsList.


paulgain:
> Looks good.
> 
> An alternative I was thinking about is having a component for each person type list item, ie:
> 
> ```
> <CardDetailsList>
>   {contacts.map(contact => <ContactListItem {...contact} />)}
> </CardDetailsList>
> ```
> 

We need to reuse the Adviser for `Added By` which is the whole point of this PR. 

This is not what we want as we would be adding a list item into table data.

**InvestmentProject.jsx and Omis.jsx**
```
<CardTable rows={[
   ...
  { header: 'Added by', content: adviser ? <AdviserListItem adviser={adviser} /> : null },
   ...
]}
``` 

Currently we are exporting this:

```
export const CardAdviser = ({ adviser }) => {
  const name = <span>{adviser.name}</span>
  const emailAddress = <Link href={`mailto:${adviser.emailAddress}`}> {adviser.emailAddress}</Link>
  const team = adviser.team ? `, (${adviser.team})` : null

  return <>{name}, {emailAddress}{team}</>
}
```

The benefit of this approach is twofold:

1. Easily wrap the CardAdviser into a list item when creating an Adviser list.
2. Reuse the CardAdviser as table data for both `InvestmentProject.jsx` and `Omis.jsx` when defining "Added By".

paulgain:
> In addition regarding the `getContacts`/`getAdvisers` you could have:
> 
> ```
> const mapPeople = (activity, mapper) => {
>   const { attributedTo } = activity.object
>   return map(filter(attributedTo, ({ type }) => {
>     return includes(type, 'dit:Contact')
>   }), mapper)
> }
> ```
> 
> Which is then used as:
> 
> ```
> const getContacts = (activity) => {
>   return mapPeople(activity, 'dit:Contact', (/* .. */) => {
>     // ...
>   })
> }
> ```

👍

PR #92, opened 2019-08-13
Add forms core

rafenden:


src/forms/core/Form.jsx
```diff
@@ -0,0 +1,65 @@
...
+  const { children, ...rest } = props
+  const formContextProps = useFormContext(rest)
+  const { submitForm } = formContextProps
+
+  const renderChildren = () => {
+    if (typeof children === 'function') {
+      return children(formContextProps)
+    }
+    return children
+  }
```
paulgain: Consider a one-liner:
`return (typeof children === 'function') ? : children(formContextProps) : children`

src/forms/journeys/MultistepExampleForm.jsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <>
+      <pre>{JSON.stringify(form, null, 2)}</pre>
+    </>
+  )
+}
+
+function sendToBackend(values) {
+  // eslint-disable-next-line no-alert
+  alert(JSON.stringify(values))
```
paulgain: Left in on purpose?

rafenden: Yes, this is just an example to help people understand how to create forms.

Will be removed once we have some forms created.

paulgain: 👍

src/forms/elements/FieldInput.jsx
```diff
@@ -0,0 +1,43 @@
...
+      <input
+        {...rest}
+        name={name}
+        key={name}
+        id={name}
+        type={type}
+        value={value}
+        onChange={onChange}
+        onBlur={onBlur}
+        style={error ? { border: 'solid 1px red' } : null}
```
paulgain: Consider importing the colour for consistency: `import { RED } from 'govuk-colours'`?

src/forms/core/__tests__/Form.test.jsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import Form from '../Form'
+import { FieldInput } from '../../elements'
+
+describe('Form', () => {
+  describe('when form has many fields', () => {
+    test('should render without issues', () => {
```
paulgain: Would be better to say `should render 14 fields without issue`

src/forms/core/useFormContext.js
```diff
@@ -0,0 +1,166 @@
+import { useState } from 'react'
+import createUseContext from 'constate'
+import { isEmpty } from 'lodash'
+
```
paulgain: Looks like a good example of how to use React Hooks. 👍

paulgain:
@rafenden looks good. I've left a few minor comments. It would be nice to see a README on how to create a Form with and without Steps despite it being clearly defined within the unit tests.

src/forms/core/Form.jsx
rafenden: I think it's slightly easier to read in the current form.

yeahfro: Just to add another option using Lodash's `isFunction` 😂 

```javascript
return isFunction(children) ? children(formContextProps) : children
```

src/forms/core/useForm.js
```diff
@@ -0,0 +1,167 @@
...
+      const newSteps = [...prevSteps]
+      newSteps.splice(currentStep + 1, 0, name)
+      return newSteps
+    }
+    return prevSteps
+  })
+  const deregisterStep = name => setSteps(prevSteps => prevSteps.filter(s => s !== name))
+
+  const goToNextStep = () => {
+    const validationErrors = validateForm()
```
yeahfro: If I am reading this step navigation correctly then it implies that we move through the steps in sequence. This is okay for now but we will need to understand that when we have more complex journeys then the user could branch in a direction.

src/forms/core/__tests__/Step.test.jsx
```diff
@@ -0,0 +1,127 @@
...
+      expect(wrapper.find('button').text()).toEqual('Next')
+      expect(formState.currentStep).toEqual(0)
+      expect(formState.steps).toEqual([
+        'testStep1',
+        'testStep2',
+      ])
+      expect(Object.keys(formState.fields)).toEqual(['testField1'])
+    })
+
+    test('should render second step when Next button is pressed', () => {
```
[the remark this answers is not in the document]
rafenden: Are you suggesting to keep `when there are more steps` as parent?

```
describe('when there are more steps')
  describe('when the Next button is pressed')
    ...
    test('...')
```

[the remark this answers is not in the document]
yeahfro: Yeah that's right

PR #93, opened 2019-08-13
Add basic entity search component

yeahfro:
## Description of change
This component is a basic version of the `Entity Search` implementation. There is a data provider which retrieves data. There is `Entity Search` component which is solely responsible for rendering the data it is given. A HOC composes the `Entity Search` component with the data provider setting the entities to be presented.

## Special attention
Reviewers should pay special attention to:
- naming of classes/functions as I was unsure regarding data providers and the `EntitySearchWithDataProvider`
- the `EntitySearchWithDataProvider` ~HOC~ hooks pattern

## Subsequent iterations
In subsequent changes there will be:
- filters
- footer with actions
- previously selected value and ability to change
- perfected styles
- negative scenarios accounted for

### Initial load
Screenshot 2019-08-14 at 14 42 16

### With results
Screenshot 2019-08-14 at 14 42 28


paulgain:
It's not a component's responsibility to make API calls - I think that's the wrong approach. It's starting to go down the same road as the Vue components (which are doing way too much) which we are moving away from. I think the data should be passed to the component therefore decoupling the component from the service call.

mcbhenwood:
I would agree that a Component ought not manage its own API interactions. In general I'd expect the scope of a Component such as this to be limited to: "I contain an input; when that input changes I call a passed callback. I contain rendering instructions for a list of items; that list of items is derived from the following state store; I re-render when that state item changes (that last bit being the job of React itself rather than the Component's code, of course)".

The job of "what happens when the callback is called and how the state alters" does not seem like the job of the component. If the component is to be reusable then its reusability would be in the general layout rules and the observation of the contracts in the previous paragraph. Where the data comes from (API / location storage / memory) and how it's filtered (API query / filter function / whatever) would be a specific of the implementation of this reusable component.

IMO, of course.

PR #95, opened 2019-08-15
feat: Introduce entity filters

yeahfro:
## Description of change
Change adds filters to the entity search. Each filter is appended as a parameter in the query string on the endpoint.

### Before
Screenshot 2019-08-15 at 21 51 07

### After
Screenshot 2019-08-15 at 21 50 55

### DIFF
--- a/src/entity-search/EntityFilter.jsx
+++ b/src/entity-search/EntityFilter.jsx
@@ -1,0 +1,27 @@
+import { InputField } from 'govuk-react'
+import { uniqueId } from 'lodash'
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const EntityFilter = ({ filter, setFilter }) => {
+  return (
+    <InputField
+      onChange={e => setFilter(filter.key, e.target.value)}
+      key={uniqueId()}
+      input={{ name: filter.key }}
+    >
+      {filter.label}
+    </InputField>
+  )
+}
+
+EntityFilter.propTypes = {
+  filter: PropTypes.shape({
+    label: PropTypes.string.isRequired,
+    key: PropTypes.string.isRequired,
+    width: PropTypes.string,
+  }).isRequired,
+  setFilter: PropTypes.func.isRequired,
+}
+
+export default EntityFilter
--- a/src/entity-search/EntityFilters.jsx
+++ b/src/entity-search/EntityFilters.jsx
@@ -1,0 +1,38 @@
+import React from 'react'
+import styled from 'styled-components'
+import { GREY_3 } from 'govuk-colours'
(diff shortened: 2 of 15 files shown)


src/entity-search/data-providers/DnbCompanySearch.js
```diff
@@ -1,9 +1,15 @@
 import axios from 'axios'
-import { compact, join } from 'lodash'
+import { compact, join, keys, map } from 'lodash'
+
+const transformObjectToQueryString = (filters) => {
+  return join(map(keys(filters), (key) => {
+    return `${encodeURIComponent(key)}=${encodeURIComponent(filters[key])}`
+  }), '&')
+}
```
paulgain: This may simply things: https://www.npmjs.com/package/query-string

src/entity-search/EntitySearch.jsx
```diff
@@ -5,20 +5,26 @@ import styled from 'styled-components'
...
   return (
     <>
+      <EntityFilters entityFilters={entityFilters} setFilter={setFilter} />
+
       {entities && entities.length ? (
         <EntityList entities={entities} />
       ) : null}
 
-      <StyledButton onClick={onEntitySearch}>Search</StyledButton>
+      <StyledButton onClick={() => onEntitySearch(filters)}>Search</StyledButton>
```
paulgain: Nice :-)

src/entity-search/EntityFilters.jsx
```diff
@@ -0,0 +1,60 @@
...
+`
+
+const StyledGridRow = styled(GridRow)`
+  & + & {
+    margin-top: ${SPACING.SCALE_2};
+  }
+`
+
+const EntityFilters = ({ entityFilters, setFilter }) => {
+  return (
```
rafenden: This return feels quite big, perhaps we could split it?

src/entity-search/EntityFilter.jsx
```diff
@@ -0,0 +1,27 @@
+import { InputField } from 'govuk-react'
+import { uniqueId } from 'lodash'
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const EntityFilter = ({ filter, setFilter }) => {
+  return (
+    <InputField
+      onInput={e => setFilter(filter.key, e.target.value)}
```
yozzo: shouldn't we use the component state for setting and getting the input value?
getting it from the dom doesn't feel like the React way...

yeahfro: @yozzo this code is using a React Hook `setFilter` to set state. The `onInput` event should probably change to `onChange`.

PR #96, opened 2019-08-15
feat: Add entity search cannot find footer

yeahfro:
## Description of change
Adds the "cannot find" footer for when the user cannot find the entity they are looking for. The footer guides them through correcting the problem.

### Before
Screenshot 2019-08-15 at 22 05 19

### After
Screenshot 2019-08-15 at 22 05 05

### DIFF
--- a/src/entity-search/EntitySearch.jsx
+++ b/src/entity-search/EntitySearch.jsx
@@ -4,6 +4,7 @@ import { Button } from 'govuk-react'
 import styled from 'styled-components'
 import { SPACING } from '@govuk-react/constants'
 
+import CannotFindDetails from './CannotFindDetails'
 import EntityList from './EntityList'
 import EntityFilters from './EntityFilters'
 import useFilter from './useFilter'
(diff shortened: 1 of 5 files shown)


src/entity-search/EntitySearch.jsx
```diff
@@ -29,6 +34,7 @@ EntitySearch.propTypes = {
     meta: PropTypes.object.isRequired,
   })).isRequired,
   onEntitySearch: PropTypes.func.isRequired,
+  cannotFind: PropTypes.object.isRequired,
```
rafenden: Could we please use `PropTypes.shape` instead?

https://stackoverflow.com/a/45764828/3042543

src/entity-search/EntitySearch.test.jsx
```diff
@@ -12,7 +13,7 @@ const API_ENDPOINT = 'http://localhost:8000/v4/dnb/company-search'
 const mock = new MockAdapter(axios)
 mock.onPost(API_ENDPOINT).reply(200, companySearchFixture)
 
-function tick() {
+function flushPromises() {
```
rafenden: Alternatively `setImmediate` might be used.

src/entity-search/EntitySearch.stories.jsx
```diff
@@ -28,6 +28,18 @@ storiesOf('EntitySearch', module)
...
               getEntities={dnbCompanySearchDataProvider(apiEndpoint)}
+              cannotFind={{
+                summary: 'I cannot find the company I am looking for',
+                actions: [
+                  'Check the country selected is correct',
+                  'Check for spelling errors in the company name',
+                  'Remove or add Ltd or Limited to your search',
+                ],
+                link: {
+                  url: 'http://blah.com',
```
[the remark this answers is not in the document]
yeahfro: Yeah good idea, I was thinking similar

src/entity-search/EntitySearch.jsx
yeahfro: Yes this is a mistake 😂

src/entity-search/EntitySearch.test.jsx
yeahfro: I tried a few combinations and the tests break.. if you know something that works better then please provide a working code example.

PR #97, opened 2019-08-16
feat: Add form date fields

paulgain:
WIP: Can you guys cast your eyes over this PR to ensure I'm on the right path. I've got lots more tests to write as well as adding more attributes like value, validation etc etc.

html

### DIFF
--- a/src/forms/elements/FieldDate.jsx
+++ b/src/forms/elements/FieldDate.jsx
@@ -1,0 +1,133 @@
+import React, { useState, useEffect } from 'react'
+import PropTypes from 'prop-types'
+import moment from 'moment'
+import { RED } from 'govuk-colours'
+import { padStart, parseInt } from 'lodash'
+
+import useFormContext from '../core/useFormContext'
+
+const DATE_FORMAT = 'YYYY-MM-DD'
+const DAY = 'day'
+const MONTH = 'month'
+const YEAR = 'year'
+
+const FieldDate = ({ name, label, validate }) => {
+  const { setFieldValue } = useFormContext()
+  const [year, setYear] = useState()
+  const [month, setMonth] = useState()
+  const [day, setDay] = useState()
+  const [errorDay, setErrorDay] = useState()
+  const [errorMonth, setErrorMonth] = useState()
+  const [errorYear, setErrorYear] = useState()
+
+  const padZero = dayOrMonth => padStart(dayOrMonth, 2, '0')
+
+  const isDay = () => (day >= 1 && day <= 31)
+  const isMonth = () => (month >= 1 && month <= 12)
+  const isYear = () => (year >= 1 && year <= 9999)
+
+  const formatDate = () => {
+    const date = moment(`${year}-${padZero(month)}-${padZero(day)}`, DATE_FORMAT, true)
+    return date.isValid() ? date.format(DATE_FORMAT) : null
+  }
+
+  const getDate = () => {
+    return {
+      date: {
+        day,
+        month,
+        year,
+      },
+      formattedDate: formatDate(),
+    }
+  }
+
+  const validateDay = () => {
+    const date = getDate()
+    if (!isDay()) {
+      date.error = DAY
+    }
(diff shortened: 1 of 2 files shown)


src/forms/elements/FieldDate.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import FieldInput from './FieldInput'
+
+const FieldDate = ({ name, legend }) => {
+  return (
+    <fieldset>
+      {legend && <legend>{legend}</legend>}
+      <FieldInput name={`${name}-day`} type="number" label="Day" pattern="[0-9]*" />
```
rafenden: I suggest to use a normal `<input />` as each field should return only one value. i.e. 
**Expected:**
```
{ myDateField: '2019-08-19' }
```
**Got:**
```
{
  myDateField-day: '19',
  myDateField-month: '08',
  myDateField-year: '2019',
}
```

src/forms/elements/FieldDate.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import FieldInput from './FieldInput'
+
+const FieldDate = ({ name, legend }) => {
```
rafenden: I suggest to use `label` instead of `legend` to be consistent with `FieldInput`.

paulgain:
@rafenden I've only just read your review. I posted this last Friday, this morning there was no response so I continued on the path I was on. In future, I suggest you give feedback earlier as this is not helpful once it's been developed.

src/forms/journeys/MultistepExampleForm.jsx
```diff
@@ -38,6 +39,13 @@ function MultistepExampleForm() {
           <Step name="first">
             <FieldInput type="number" name="age" label="Age" validate={value => (!value ? 'Age is required' : null)} />
             <FieldInput type="number" name="height" label="Height" validate={value => (!value ? 'Height is required' : null)} />
+            <FieldDate
+              name="dob"
+              legend="Date of birth"
+              validateDay={value => (!value ? 'Day is required' : null)}
```
rafenden: I suggest having a single validation callback and then pass the complete date as value, same as `FieldInput`.

src/forms/elements/FieldDate.jsx
paulgain: 👍

src/forms/journeys/MultistepExampleForm.jsx
paulgain: 👍

src/forms/elements/FieldDate.jsx
paulgain: 👍

paulgain:
Currently form validation is performed when the user clicks next. I'm  unsure if this is the best approach considering we're using React - we have everything we need within the browser. The current approach seems to mimic our existing implementation of Node/Express/Nunjucks where
the validation is done server side when the user clicks next - I don't know if this is intentional.

 We could improve client side validation by validating a field once it has lost focus rather than waiting for the user to complete the entire form and then informing them where they've gone wrong. I don't know how much effort it would be to make those changes, we're at an early stage so it should be minimal.

I've applied validation to each of the day fields (day, month and year). The validation is applied `onBlur` (element has lost focus) therefore the user receives immediate feedback as just mentioned.

web-bert:
GDS to not recommend immediate validation:

> Only display an error when someone tries to move to the next part of the service. Do not show error messages:
> * when users select or tab to a field
> * as they are typing
> * when they move away from a field

https://design-system.service.gov.uk/components/error-message/

paulgain:
> GDS do not recommend immediate validation:
> 
> > Only display an error when someone tries to move to the next part of the service. Do not show error messages:
> > 
> > * when users select or tab to a field
> > * as they are typing
> > * when they move away from a field
> 
> https://design-system.service.gov.uk/components/error-message/

👍 

GDS also

PR #100, opened 2019-08-19
feat: Use govuk-react buttons for navigation across the form

rafenden:
https://trello.com/c/VmYAQUXx/146-265-deploy-a-skeleton-of-the-frontend-journey-behind-a-feature-flag-fe

image


src/forms/core/Step.jsx
```diff
@@ -1,13 +1,15 @@
 import React, { useEffect } from 'react'
 import PropTypes from 'prop-types'
+import { Button } from 'govuk-react'
+import { BLACK, GREY_2 } from 'govuk-colours'
 
 import useFormContext from './useFormContext'
 
-function Step({ name, children }) {
+function Step({
+  name, backButtonText, forwardButtonText, hideBackButton, hideForwardButton, children }) {
```
yeahfro: This definition feels like it could be broken over multiple lines.

src/forms/core/Step.jsx
```diff
@@ -26,26 +28,46 @@ function Step({ name, children }) {
     return null
   }
 
+  const defaultBackButtonText = 'Back'
```
yeahfro: Could `backButtonText` just be given a default value in the function signature?

src/forms/core/Step.jsx
yeahfro: I am wondering if `hideBackButton` and `hideForwardButton` are required. ie if `null` is passed for those values does it implicitly mean the button can be hidden?

src/forms/core/Step.jsx
rafenden: Unfortunately, I can't as in the case of `defaultForwardButtonText` it's dynamic based on condition.

src/forms/core/Step.jsx
rafenden: PropType is `boolean` so you would get an error if null was passed.

yeahfro: Okay I would say it can be useful to have a tri state bool which effectively evaluates to `true`, `false` and `not set`. I would say that is cleaner than introducing another parameter to achieve the same thing.

PR #106, opened 2019-08-19
Add search result onClick event to entity search component

yeahfro:
## Description of change
Adds the entity search result click event to the entity search component.

## Test instructions
1. Browse to `/?path=/story/entitysearch--data-hub-company-search`
2. Click `Search`
3. Click a search result. You should see an `alert`.

## Screenshots
There should be no visual change although in Storybook the click event will `alert` the entity as a string.

### DIFF
--- a/src/entity-search/EntityList.jsx
+++ b/src/entity-search/EntityList.jsx
@@ -30,6 +30,7 @@ const StyledEntity = styled('div')`
     }
   }
 `
+StyledEntity.displayName = 'StyledEntity'
 
 const StyledHeading = styled(H3)`
   margin: 0;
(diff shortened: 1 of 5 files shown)


src/entity-search/EntityList.jsx
```diff
@@ -21,6 +21,7 @@ const StyledEntity = styled('div')`
   border: 1px solid ${GREY_3};  
   cursor: pointer;
 `
+StyledEntity.displayName = 'StyledEntity'
```
stackzak: I hadn't come across this before so I looked it up and found that cccording to the MDN webdocs:
> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

src/entity-search/EntityList.jsx
```diff
@@ -36,19 +37,19 @@ const StyledMetaItem = styled('div')`
   }
 `
 
-const EntityList = ({ entities }) => {
+const EntityList = ({ entities, onEntityClick }) => {
   return (
     <StyledEntityList>
-      {entities.map(({ heading, meta }) => {
+      {entities.map((entity) => {
         return (
```
stackzak: Not a required change but you don't need the return statements and curly brackets when using arrow functions i.e
from: `something.map(thing => { return (<component />) }`
to: `something.map(thing => (<component />))`

src/entity-search/EntityList.jsx
yeahfro: @stackzak this is a recommended approach to testing Styled Components with Enzyme. It is being used for testing and not a dependency of the code running in production.

PR #111, opened 2019-08-22
feat: Show message when no entities returned

yeahfro:
## Description of change
When no entities are returned from the API request, a friendly message should be presented to the user rather than nothing at all.

## Test instructions
1. Browse to `/?path=/story/entitysearch--data-hub-company-search-with-no-results`
2. Click `Search`. You should see `There are no entities to show.`

## Screenshot
Screenshot 2019-08-22 at 14 01 09

### DIFF
--- a/src/entity-search/EntitySearch.jsx
+++ b/src/entity-search/EntitySearch.jsx
@@ -38,6 +38,10 @@ const EntitySearch = ({
         </>
       ) : null}
 
+      {entities && entities.length === 0 ? (
+        <p>There are no entities to show.</p>
+      ) : null}
+
       {error && <p>{error}</p>}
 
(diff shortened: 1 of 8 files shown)


src/entity-search/EntitySearch.jsx
```diff
@@ -38,6 +38,10 @@ const EntitySearch = ({
         </>
       ) : null}
 
+      {entities && entities.length === 0 ? (
+        <p>There are no entities to show.</p>
```
rafenden: Shall we make this text to stand out more?
Not sure if it's visible enough.

yeahfro: I think that's a good point. I will bounce it off the designers. For now I will merge this and possibly follow up with style tweaks.

PR #112, opened 2019-08-22
Present text when entity is known, handle click event differently

yeahfro:
## Description of change
For some entities we may want to display some inset text and therefore handle click events in a different way.

An example of this is for companies that are known by DIT. We will display a message containing a link so that the user can browse to the company page and add activity. In this case the entity click event should be ignored. When the company is not known by DIT then the entity click event is handled to move to the next stage of the process for adding a company.

## Test instructions
1. Browse to `/?path=/story/entitysearch--data-hub-company-search`
2. Click the first entity and the click event is ignored. Click the `Go to the company page` link the URL will change
3. Click the second entity and the click event is handled with an `alert`

### Before
Screenshot 2019-08-22 at 14 59 48

### After
Screenshot 2019-08-22 at 14 59 07

### DIFF
--- a/src/entity-search/EntityList.jsx
+++ b/src/entity-search/EntityList.jsx
@@ -2,9 +2,15 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { SPACING, FONT_SIZE, MEDIA_QUERIES } from '@govuk-react/constants'
-import { GREY_2, GREY_4, LINK_COLOUR, LINK_HOVER_COLOUR, SECONDARY_TEXT_COLOUR } from 'govuk-colours'
+import {
+  GREY_2,
+  GREY_4,
+  LINK_COLOUR,
+  LINK_HOVER_COLOUR,
+  SECONDARY_TEXT_COLOUR,
+} from 'govuk-colours'
 import { uniqueId } from 'lodash'
-import { H3 } from 'govuk-react'
+import { H3, InsetText } from 'govuk-react'
 
 const StyledEntityList = styled('ol')`
   margin: ${SPACING.SCALE_2} 0 ${SPACING.SCALE_4} 0;
@@ -21,14 +27,16 @@ const StyledEntity = styled('div')`
   border: 1px solid ${GREY_2};  
   cursor: pointer;
   
-  &:hover {
-    border: 1px solid ${LINK_HOVER_COLOUR};
-    background-color: ${GREY_4};
-    
-    & > h3 {
-      color: ${LINK_HOVER_COLOUR};
+  ${({ canHandleClick }) => canHandleClick && `
+    &:hover {
+      border: 1px solid ${LINK_HOVER_COLOUR};
+      background-color: ${GREY_4};
+      
+      & > h3 {
+        color: ${LINK_HOVER_COLOUR};
+      }
     }
-  }
+  `}
 `
 StyledEntity.displayName = 'StyledEntity'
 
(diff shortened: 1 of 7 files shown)


src/entity-search/EntityList.jsx
```diff
@@ -52,20 +60,29 @@ const StyledMetaItem = styled('div')`
   }
 `
 
+const StyledInsetText = styled(InsetText)`
+  margin-top: ${SPACING.SCALE_2} !important;
```
yozzo: I'm still yet to be convinced by the benefits of this library

src/entity-search/EntitySearch.stories.jsx
```diff
@@ -56,7 +56,11 @@ const EntitySearchForStorybook = ({ previouslySelected, cannotFindLink }) => {
               ],
               link: cannotFindLink,
             }}
-            onEntityClick={entity => alert(`Selected ${JSON.stringify(entity)}`)}
+            onEntityClick={(entity) => {
+              if (!entity.datahub_company) {
+                alert(`Selected ${JSON.stringify(entity)}`)
```
yozzo: is this an interim thing, or we're supposed to go to prod with `alert`?

yeahfro: This is a Storybook story. It is demonstrating the use of a callback.

src/entity-search/EntityList.jsx
yeahfro: I don't like this either - open to ideas.

stackzak: Are we referring to Styled Components? If so the benefit is that it's in alignment with what's being used in the gov-uk-react repo which let's us override their styles when needed (which is often).

yeahfro: [I've replaced this line with something a bit more SC style](https://github.com/uktrade/data-hub-components/pull/112/files#diff-45112ab8a73303f705ca4f4c18eb6c3cR41)

src/entity-search/EntityList.jsx
```diff
@@ -2,9 +2,15 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { SPACING, FONT_SIZE, MEDIA_QUERIES } from '@govuk-react/constants'
-import { GREY_2, GREY_4, LINK_COLOUR, LINK_HOVER_COLOUR, SECONDARY_TEXT_COLOUR } from 'govuk-colours'
+import {
+  GREY_2,
```
mcbhenwood: Should probably put reformatting into a separate commit, however feel free to ignore this as I see that in the following commit these imports are removed from this module.

src/entity-search/data-providers/DnbCompanySearch.jsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable camelcase */
+import React from 'react'
```
mcbhenwood: Oughtn't we observe alphabetical sorting here?

src/entity-search/EntityListEntity.jsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+`
+
+const StyledInsetText = styled(InsetText)`
+  & {
+    margin-top: ${SPACING.SCALE_2};
+  }
+`
+
+const EntityListEntity = ({ canHandleClick, onEntityClick, data, text, heading, meta }) => {
```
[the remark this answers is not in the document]
yeahfro: Yeah I didn't like the naming either, glad you picked it up. I will got with EntityListItem as it probably reads better although I have some discomfort around it not technically being a List Item.

src/entity-search/data-providers/DnbCompanySearch.jsx
yeahfro: If we plan to do that then let's just bounce it off other developers as we have not done it so far. Happy to consider it but perhaps should update the repo and lint rules at the same time.

PR #116, opened 2019-08-27
feat: Support multiple validators in forms field

rafenden:
Let users define more than one validator in fields.

Before:
```
<FieldInput
  type="text"
  name="name"
  label="Name"
  validate={value => (!value ? 'name is required' : null)}
/>
```

After:
```
<FieldInput
  type="text"
  name="name"
  label="Name"
  validate=[
    validateRequired,
    validateName,
  ]
/>
```

### DIFF
--- a/src/forms/core/__tests__/useForm.test.jsx
+++ b/src/forms/core/__tests__/useForm.test.jsx
@@ -3,25 +3,67 @@ import { renderHook, act } from '@testing-library/react-hooks'
 import useForm from '../useForm'
 
 const validate = value => (value !== 'correctValue' ? 'testError1' : null)
-const testFields = [
-  {
-    name: 'testField1',
-    label: 'testLabel1',
-    initialValue: 'testInitialValue1',
-    validate,
-  },
-  {
-    name: 'testField2',
-    label: 'testLabel2',
-  },
-]
+const testField1 = {
+  name: 'testField1',
+  label: 'testLabel1',
+  initialValue: 'testInitialValue1',
+  validate,
+}
+const testField2 = {
+  name: 'testField2',
+  label: 'testLabel2',
+}
 
 describe('useForm', () => {
+  describe('when getFieldState() is called', () => {
+    let fieldState
+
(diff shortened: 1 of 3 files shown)


src/forms/core/useForm.js
```diff
@@ -20,11 +20,16 @@ function useForm({ initialValues = {}, initialStep = 0, onSubmit = null } = {})
     const value = values[name]
 
     if (!field) {
-      return 'Field does not exists'
+      return `Field ${name} does not exists`
```
yeahfro: `exists` 👉 `exist`

src/forms/core/useForm.js
```diff
@@ -20,11 +20,16 @@ function useForm({ initialValues = {}, initialStep = 0, onSubmit = null } = {})
...
     }
 
-    if (field && 'validate' in field && typeof field.validate === 'function') {
-      return field.validate(value, name)
+    if (field && 'validate' in field) {
+      if (field.validate instanceof Function) {
+        return field.validate(value, name)
+      } else if (field.validate instanceof Array) {
```
yeahfro: The `else if` is not really required here, ie:

```
if (field.validate instanceof Function) {
  return field.validate(value, name)
}

if (field.validate instanceof Array) {
  const error = field.validate.map(validator => validator(value, field))
  return error instanceof Array ? error[0] : error
}
```

src/forms/core/__tests__/useForm.test.jsx
```diff
@@ -16,6 +16,12 @@ const testFields = [
   },
 ]
 
+const flushPromises = () => {
```
yeahfro: We should probably move this out into somewhere common.

rafenden: https://github.com/testing-library/react-testing-library has this function built in.

rafenden: `flushPromises` was replaced by `await Promise.resolve()`

PR #117, opened 2019-08-27
fix: Set key to a non changing ID

yeahfro:
## Description of change
When the component was being rendered, it was being given a new key with uniqueId which was
causing bad effects in React. In this particular case the text field was losing focus and could not be updated.

## Test instructions
1. Browse to https://uktrade.github.io/data-hub-components/?path=/story/entitysearch--data-hub-company-search
2. Type in the filters, the filter will lose focus


src/entity-search/EntityFilters.jsx
```diff
@@ -3,20 +3,40 @@ import styled from 'styled-components'
...
+}
+
 const EntityFilters = ({ entityFilters, setFilter }) => {
   return (
     <StyledFilters>
-      {entityFilters.map(filterRow => (
-        <EntityFilterRow filterRow={filterRow} setFilter={setFilter} key={uniqueId()} />
+      {entityFilters.map((filterRow, id) => (
+        <StyledGridRow key={getKey('grid_row', id)}>
+          {filterRow.map(filter => (
```
ian-leggett: Really small, is it worth destructuring `{width, key}` here?

yeahfro: @ian-leggett `filter` is required a few lines down. An option could be to do it on the following line ie `const { width, key } = filter` but it doesn't really add much and arguably loses readability.

ian-leggett: Good point, all good with me 👍

PR #129, opened 2019-08-29
Fix results and errors continually showing when state changes for subsequent "Search" clicks

yeahfro:
## Description of change
This change addresses two similar issues:
1. When the user clicks "Search" and there is an error and then clicks "Search" again and there are results then both error and results are being displayed. In this scenario, the error is hidden and the results are shown.
2. Same as above but results first and error second.

In addition, the use of mocks has been refactored as it was being used in multiple places.

### Example
Screenshot 2019-08-29 at 11 10 08


src/entity-search/EntitySearch.test.jsx
```diff
@@ -1,24 +1,13 @@
 import React from 'react'
 import { mount } from 'enzyme'
-import MockAdapter from 'axios-mock-adapter'
-import axios from 'axios'
 import { act } from 'react-dom/test-utils'
 
-import fixtures from './fixtures'
+import { setupSuccessMocks, setupErrorMocks, setupNoResultsMocks } from './__tests__/mocks'
+import fixtures from './__tests__/fixtures'
```
rafenden: I suggest to place fixtures in `./src/entity-search/__fixtures__` as they might be also used in stories and not only in tests.

src/entity-search/EntitySearch.stories.jsx
```diff
@@ -2,48 +2,15 @@ import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { GridCol, GridRow, Main, H2 } from 'govuk-react'
 import { SPACING } from '@govuk-react/constants'
-import axios from 'axios'
-import MockAdapter from 'axios-mock-adapter'
 import PropTypes from 'prop-types'
 
+import { setupSuccessMocks, setupErrorMocks, setupNoResultsMocks } from './__tests__/mocks'
```
rafenden: Did you consider using Jest manual mocks? They could be placed in `/src/entity-search/__mocks__`.
See the docs: https://jestjs.io/docs/en/manual-mocks.html

src/entity-search/EntitySearch.test.jsx
yeahfro: Yeah nice one, I've done that now

src/entity-search/EntitySearch.stories.jsx
yeahfro: Okay I have considered this a couple of times and having experimented with it, the manual mocks is a nice feature but is not very reusable with Storybook. This is because of the `jest` dependency so there does not appear to be a clean way of doing this. I am sticking with the current method for now. I have moved the mocks to a `__mocks__` directory.

PR #131, opened 2019-08-29
feat: Validate form only on submit

rafenden:
GDS does not recommend validating forms using `onBlur` event, the PR will remove this behaviour:
https://design-system.service.gov.uk/components/error-message/

Thanks to @web-bert for pointing that out:
https://github.com/uktrade/data-hub-components/pull/97#issuecomment-524239998


src/forms/core/__tests__/useField.test.jsx
```diff
@@ -29,38 +30,58 @@ const TestField = (props) => {
 }
 
 describe('useField', () => {
-  describe('when text is typed to the field', () => {
-    test('should update state value', () => {
-      const wrapper = mount(
+  describe('when field is mounted with `initialValue`', () => {
```
yeahfro: 👍 these tests are looking quite tight

src/forms/core/useForm.js
```diff
@@ -26,11 +30,11 @@ function useForm({ initialValues = {}, initialStep = 0, onSubmit = null } = {})
     }
 
     if (field && 'validate' in field) {
-      if (field.validate instanceof Function) {
+      if (typeof field.validate === 'function') {
```
yeahfro: Based on using `isEmpty` from `lodash` I wonder if you should use `isFunction`

src/forms/core/useForm.js
```diff
@@ -26,11 +30,11 @@ function useForm({ initialValues = {}, initialStep = 0, onSubmit = null } = {})
...
         return field.validate(value, name)
       }
 
-      if (field.validate instanceof Array) {
+      if (Array.isArray(field.validate)) {
```
yeahfro: And `isArray`

src/forms/core/useForm.js
rafenden: As a rule of thumb, I prefer using built-in functions where possible to make the bundle smaller.

PR #137, opened 2019-09-02
fix: Use checkbox instead of link to expand activities

rafenden:
https://trello.com/c/G0RJOMn2/233-use-checkbox-in-activity-feed-to-show-details-for-all-cards

## Before
image

## After
image

### DIFF
--- a/src/activity-feed/__tests__/ActivityFeedFilters.test.jsx
+++ b/src/activity-feed/__tests__/ActivityFeedFilters.test.jsx
@@ -23,24 +23,21 @@ describe('ActivityFeedFilters', () => {
     })
   })
 
-  describe('when the "Show details for all activities" link is clicked', () => {
-    let onShowDetailsClickMock
-    let preventDefaultMock
+  describe('when the "Show details for all activities" checkbox is clicked', () => {
+    const onShowDetailsClickMock = jest.fn()
 
     beforeEach(() => {
-      onShowDetailsClickMock = jest.fn()
-      preventDefaultMock = jest.fn()
-
       const wrapper = shallow((<ActivityFeedFilters
         onShowDetailsClick={onShowDetailsClickMock}
         showDetails={false}
       />))
-
-      wrapper.find('ShowDetails').simulate('click', { preventDefault: preventDefaultMock })
+      wrapper.find('Checkbox').simulate('change', {
+        target: { checked: true },
+      })
     })
 
-    test('should prevent default', () => {
-      expect(preventDefaultMock.mock.calls.length).toEqual(1)
+    afterEach(() => {
+      onShowDetailsClickMock.mockReset()
     })
 
     test('should call onShowDetailsClick', () => {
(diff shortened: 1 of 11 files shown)


src/activity-feed/cards/card/CardTable.jsx
```diff
@@ -36,7 +36,7 @@ class DetailsRow extends React.PureComponent {
 
     return (
       <Table.Row>
-        <Table.CellHeader style={{ fontWeight: 'normal', border: 0 }}>{header}</Table.CellHeader>
+        <Table.CellHeader style={{ fontWeight: 'normal', border: 0, width: '35%' }}>{header}</Table.CellHeader>
```
mcbhenwood: Is it always 35%?

rafenden:
> What happens if the user then unchecks the checkbox? It feels like a bit of an open-ended journey at the moment.

Here's a video of this functionality:
image

src/activity-feed/cards/card/CardTable.jsx
rafenden: Here's a comparison of CardTable.

## With fixed column width:
image

## With dynamic column width:
image

src/activity-feed/__tests__/ActivityFeedFilters.test.jsx
```diff
@@ -23,24 +23,21 @@ describe('ActivityFeedFilters', () => {
     })
   })
 
-  describe('when the "Show details for all activities" link is clicked', () => {
-    let onShowDetailsClickMock
-    let preventDefaultMock
+  describe('when the "Show details for all activities" checkbox is clicked', () => {
+    const onShowDetailsClickMock = jest.fn()
```
yeahfro: Nice, I like the setting and reset of `onShowDetailsClickMock`

src/forms/elements/Form.jsx
```diff
@@ -68,4 +37,35 @@ FormWrapper.defaultProps = {
   children: null,
 }
 
+function Form(props) {
```
yeahfro: How come this is here?

rafenden: I just changed the order.

yeahfro: It doesn't make much sense to have this as part of the same commit as it is unrelated.

rafenden: Agree, this change is now discarded.

mcbhenwood:
@rafenden Thanks for the video. I don't see any tests for the collapse on the second click.

src/activity-feed/cards/card/CardTable.jsx
ian-leggett: Should we be inline styling here? I thought we would override with styled components.

debitan: I have a similar thought. Looks like we're using styled components for all our other styling needs - so seems out of place to have inline styling here. I'm guessing you might have a reason as to why though?

rafenden: This change has been reverted as it was not related to the main purpose of this PR. It will be coming back in a new PR.

rafenden:
> What happens if the user then unchecks the checkbox? It feels like a bit of an open-ended journey at the moment.

Test added for the `unchecked` state.

PR #138, opened 2019-09-02
feat: Add default exports for components

yeahfro:
## Description of change
These are missing and useful for consuming the library.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -11,15 +11,8 @@ export { default as MyCompaniesTile } from './dashboard/my-companies/MyCompanies
 export { default as useMyCompaniesContext } from './dashboard/my-companies/useMyCompaniesContext'
 
 // Entity search
-export { default as CannotFindDetails } from './entity-search/CannotFindDetails'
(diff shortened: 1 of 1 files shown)


src/index.jsx
```diff
@@ -10,12 +10,25 @@ export { default as ActivityFeedPagination } from './activity-feed/ActivityFeedP
...
 
+// Entity search
+export { default as CannotFindDetails } from './entity-search/CannotFindDetails'
+export { default as EntityFilter } from './entity-search/EntityFilter'
+export { default as EntityFilters } from './entity-search/EntityFilters'
+export { default as EntityList } from './entity-search/EntityList'
+export { default as EntityListItem } from './entity-search/EntityListItem'
+export { default as EntityListItemMetaList } from './entity-search/EntityListItemMetaList'
+export { default as EntitySearch } from './entity-search/EntitySearch'
+export { default as EntitySearchWithDataProvider } from './entity-search/EntitySearchWithDataProvider'
```
rafenden: Not sure if we need to export all of these?

PR #140, opened 2019-09-06
refactor: Remove dependency on snapshots

yeahfro:
## Description of change
We are migrating away from snapshot testing. This change removes the dependency on snapshots for testing of the entity search component.


src/entity-search/__tests__/EntitySearch.test.jsx
```diff
@@ -56,36 +56,51 @@ describe('EntitySearch', () => {
...
-    beforeAll(async () => {
-      setupSuccessMocks(API_ENDPOINT)
+    test('should not show the entities', () => {
+      expect(wrappedEntitySearch.find('ol').exists()).toBeFalsy()
+    })
 
-      wrappedEntitySearch = wrapEntitySearch()
-      preventDefaultMock = jest.fn()
+    test('should not show the cannot find details', () => {
+      expect(wrappedEntitySearch.find('ul').exists()).toBeFalsy()
```
rafenden: In theory, you could also try checking for element existence using the component itself, i.e.
```
expect(wrappedEntitySearch.find(CannotFindDetails).exists()).toBeFalsy()
```
Not sure though if this will work in all cases.

mcbhenwood:
Agreed. As discussed recently. snapshot testing is not a good fit in the case of this repo and its other testing disciplines.

src/entity-search/__tests__/EntitySearch.test.jsx
yeahfro: @rafenden nice, I have done this and will also use it from now on.

PR #142, opened 2019-09-09
Add StatusMessage to EntitySearch

yeahfro:
## Description of change
This change introduces a status message component. This is a common pattern used across Data Hub. In this particular instance it is required for displaying information at the top of the entity search component.

This change includes:
- `StatusMessage` component
- integrating component with existing entity search
- integrating component with `FieldDnbCompany`

## Test instructions
Browse to any:
- `/?path=/story/statusmessage--default`
- `/?path=/story/statusmessage--custom-colour`
- `/?path=/story/entitysearch--data-hub-company-search-with-entity-list-header`, click 
- `/?path=/story/forms--fielddnbcompany-default`, click `Search`

### StatusMessage
Screenshot 2019-09-09 at 16 34 57
Screenshot 2019-09-09 at 15 31 51

### EntitySearch
Screenshot 2019-09-09 at 15 32 31

### Field DnbCompany
Screenshot 2019-09-09 at 15 32 21

PR #145, opened 2019-09-10
feat: Add DetailsText component

reupen:
This is similar to InsetText, but with a a grey background.

It is a rough port of the `Message({type: 'muted'})` Nunjucks macro from data-hub-frontend.

## To test

Navigate to `DetailsText` in Storybook.

## Screenshot

image


src/details-text/DetailsText.jsx
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components'
+import InsetText from '@govuk-react/inset-text'
+import { GREY_4 } from 'govuk-colours'
+
+
+const DetailsText = styled(InsetText)`
```
rafenden: Perhaps, we could name it `InsetText` so it's not confused with the `Details` component?

reupen: Yes, we could find a better name. I'm not sure about putting 'muted' in the name though, it's intended to draw attention to what's contained in it.

rafenden: Or even better, it could be merged with the new `StatusMessage` component added in https://github.com/uktrade/data-hub-components/pull/142
@yeahfro what's your opinion on this?

src/details-text/DetailsText.jsx
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components'
```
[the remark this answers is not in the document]
reupen: Done.

src/details-text/DetailsText.jsx
yeahfro: I am wondering if this component is required as maybe it should just be an `InsetText` without the grey background. Maybe this is a conversation with design?

yeahfro: We have been naming styled components with the prefix `Styled`, so in this instance `StyledInsetText` or `StyledDetailsText`.

reupen: I queried it already ;) It's already being used in Data Hub is the short answer.

reupen:
Closing as we agreed to use `InsetText` as it is.

PR #146, opened 2019-09-10
feat: Add FormActions component

reupen:
This is a port of the `c-form-actions` CSS class from data-hub-frontend.

The same styling has been used.

## To test

Navigate to `Forms/FormActions` in Storybook.

## Screenshot

image


src/forms/elements/FormActions.jsx
```diff
@@ -0,0 +1,25 @@
+import styled from 'styled-components'
+import { MEDIA_QUERIES, SPACING } from '@govuk-react/constants'
+
+const FormActions = styled.div`
```
yeahfro: Not an issue but in most places we are doing `styled('div')`

src/forms/elements/FormActions.jsx
```diff
@@ -0,0 +1,25 @@
...
+  * {
+    display: block;
```
yeahfro: I don't think this is required. It's implicit, right?

reupen: Good question. For the current class in Data Hub it makes a difference on mobile. However, it doesn't seem to have much effect here. I'll remove it and we can add it back if we find a reason to.

PR #152, opened 2019-09-12
New form for creating new lists for companies

ian-leggett:
## Description
This is the early stage for creating new company lists. When following a company that has not been added to a list you will see this form which enables you to create a new list.

Screenshot 2019-09-12 at 09 37 02


src/company-lists/CreateListForm.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import Button from '@govuk-react/button'
+import PropTypes from 'prop-types'
+import Form from '../forms/elements/Form'
+import FieldInput from '../forms/elements/FieldInput'
+
+const CreateListForm = (props) => {
+  const { onSubmitHandler } = props
+  const { id, name, label, maxLength } = props
```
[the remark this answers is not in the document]
ian-leggett: Good idea 👍

src/company-lists/CreateListForm.jsx
```diff
@@ -0,0 +1,51 @@
...
+const CreateListForm = (
+  { id,
+    name,
+    label,
+    hint,
+    maxLength,
+    onSubmitHandler,
+  },
+) => {
+  const companyLink = `/companies/${id}`
```
reupen: Would suggest that we pass a `returnUrl` to the component (rather than build a URL for the company here) here so this component doesn't need to know what the front end URLs are, and also so the component isn't tied to that particular page.

src/company-lists/CreateListForm.jsx
```diff
@@ -0,0 +1,51 @@
...
+        required="Enter a name for your list"
+        hint={hint}
+        validate={value => (value && value.length > maxLength
+          ? `Enter list name which is no longer than ${maxLength} characters`
+          : null)
+        }
+      />
+      <Button>Create list</Button>
+      <div>
+        <a href={companyLink}>Cancel</a>
```
reupen: Should we be using [the `Link` component from govuk-react](https://govuk-react.github.io/govuk-react/?path=/story/typography-link--component-default)?

src/company-lists/CreateListForm.jsx
ian-leggett: 👍

src/company-lists/CreateListForm.jsx
ian-leggett: 👍

src/index.jsx
```diff
@@ -34,3 +34,6 @@ export { default as FieldSelect } from './forms/elements/FieldSelect'
 
 // Status message
 export { default as StatusMessage } from './status-message/StatusMessage'
+
+// Company lists
+export { default as CreateListForm } from './company-lists/CreateListForm'
```
reupen: Would be good to move this into the company lists section further up.

PR #155, opened 2019-09-13
Add address search

paulgain:
## Test
Go to `Address Search` within Storybook, open `knobs` by pressing `a`. Here you will need to add two strings:
1. A base URL
2. API key for the postcode lookup URL (vault)

Once these are in place you will be able to lookup a UK address.

## Animated GIF
address-lookup


src/address-search/data-providers/AddressSearch.js
```diff
@@ -0,0 +1,31 @@
...
+    county: isEmpty(addressParts[5]) || isEmpty(addressParts[6]) ? null : addressParts[6],
+    city: isEmpty(addressParts[5]) ? addressParts[6] : addressParts[5],
+    address1: isEmpty(addressParts[2]) ? addressParts[0] : `${addressParts[0]}${'-'}${addressParts[1]}`,
+    address2: isEmpty(addressParts[2]) ? addressParts[1] : addressParts[2],
+  }, value => !isNil(value) && !isEmpty(value))
+
+  return {
+    ...parsed,
+  }
+}
```
paulgain: Taken directly from Data Hub: https://github.com/uktrade/data-hub-frontend/blob/develop/src/apps/addresses/services/lookup.js#L15-L30

This can be improved once we have more time.

src/address-search/__stories__/AddressSearch.stories.jsx
```diff
@@ -0,0 +1,20 @@
...
+import { withKnobs, text } from '@storybook/addon-knobs'
+
+import addressSearch from '../data-providers/AddressSearch'
+import AddressSearchWithDataProvider from '../AddressSearchWithDataProvider'
+
+addDecorator(withKnobs)
+
+storiesOf('Address search', module)
+  .add('Data Hub address search', () => {
+    const baseUrl = text('BASE URL', 'please add the baseUrl here')
```
rafenden: I suggest using the real URL here, it's an external service so it should be OK to provide the URL in examples

src/address-search/__stories__/AddressSearch.stories.jsx
```diff
@@ -0,0 +1,20 @@
...
+    const apiKey = text('API KEY', 'please add the API key here')
```
rafenden: I suggest to use `YOUR_GETADDRESS_IO_API_KEY`

src/address-search/AddressSearch.jsx
```diff
@@ -0,0 +1,57 @@
...
+          error={error}
+          onChange={evt => setPostCode(evt.target.value)}
+        />
+        <Button onClick={onAddressSearchClick}>Find UK Address</Button>
+      </div>
+      { addressList && (
+        <Select name="address" label="Select an address">
+          {
+            addressList.map((value, index) => {
+              return <option key={uniqueId()} value={index}>{ value.address1 }</option>
```
rafenden: We could use the `value.address1` as the `key`.

src/address-search/__tests__/AddressSearch.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+  describe('when the component initially loads', () => {
+    let wrapper
+
+    beforeAll(() => {
+      setupSuccessMocks(GOOD_POSTCODE_URL)
+      wrapper = getWrapper()
+    })
+
+    test('should render the component', () => {
+      expect(wrapper).toMatchSnapshot()
```
rafenden: Please, do not use snapshot tests, as they are hard to read.

src/address-search/AddressSearch.jsx
paulgain: I thought that as well, however, it doesn't turn out how you would think. If you search for the postcode `BN1 4SE`
(you can see the results further down the page) you will see there are a number of flats with the same 1st line: For example:

```
 1.  "Flat 1, 109 Ditchling Road, , , , Brighton, East Sussex",
 2.  "Flat 1, 119 Ditchling Road, , , , Brighton, East Sussex",
 3.  "Flat 1, 137 Ditchling Road, , , , Brighton, East Sussex",
```

Currently in Data Hub we only show the first line, if we were to use that as a `key` we would have duplicate keys.

src/address-search/__tests__/AddressSearch.test.jsx
paulgain: I've literally followed suit for how the `Entity Search` component has been tested, PR'd and merged.

src/address-search/__stories__/AddressSearch.stories.jsx
paulgain: 👍

src/address-search/__stories__/AddressSearch.stories.jsx
paulgain: Shouldn't we use this instead: https://github.com/uktrade/data-hub-frontend/blob/develop/sample.env#L1?

paulgain:
> Well done!
> A few minor comments.

👍

src/address-search/AddressSearch.jsx
yeahfro: Could we generate an ID at API request time which is the full address? Then that will be accessible here. In short there are issues with using `uniqueId` as the `key` which we need to refactor out of our other code.

src/address-search/__tests__/AddressSearch.test.jsx
yeahfro: Entity search no longer uses snapshot testing - maybe you have an old version?

src/address-search/data-providers/AddressSearch.js
```diff
@@ -0,0 +1,31 @@
...
+  const parsed = pickBy({
+    postcode,
+    county: isEmpty(addressParts[5]) || isEmpty(addressParts[6]) ? null : addressParts[6],
+    city: isEmpty(addressParts[5]) ? addressParts[6] : addressParts[5],
+    address1: isEmpty(addressParts[2]) ? addressParts[0] : `${addressParts[0]}${'-'}${addressParts[1]}`,
+    address2: isEmpty(addressParts[2]) ? addressParts[1] : addressParts[2],
+  }, value => !isNil(value) && !isEmpty(value))
+
+  return {
+    ...parsed,
```
yeahfro: You could just `return parsed`

src/address-search/data-providers/AddressSearch.js
```diff
@@ -0,0 +1,31 @@
...
+import axios from 'axios'
+import {
+  map,
+  trim,
+  pickBy,
+  isEmpty,
+  isNil,
+} from 'lodash'
+
+const transformAddress = (address, postcode) => {
```
yeahfro: Would be good to also copy across the tests as it's rather hairy.

paulgain: 👍

src/address-search/AddressSearch.jsx
paulgain: 👍

src/address-search/__tests__/AddressSearch.test.jsx
paulgain: I'll pull these out within the next PR is that's OK?

src/address-search/data-providers/AddressSearch.js
```diff
@@ -6,20 +6,33 @@ import {
...
-    city: isEmpty(addressParts[5]) ? addressParts[6] : addressParts[5],
-    address1: isEmpty(addressParts[2]) ? addressParts[0] : `${addressParts[0]}${'-'}${addressParts[1]}`,
+    townOrCity: isEmpty(addressParts[5]) ? addressParts[6] : addressParts[5],
+    address1: isEmpty(addressParts[2]) ? addressParts[0] : `${addressParts[0]}${' - '}${addressParts[1]}`,
     address2: isEmpty(addressParts[2]) ? addressParts[1] : addressParts[2],
   }, value => !isNil(value) && !isEmpty(value))
 
   return {
     ...parsed,
+    id: addressObjToString(parsed),
```
yeahfro: Could the `id` just have been the original `address` parameter?

src/address-search/__tests__/AddressSearch.test.jsx
yeahfro: That's okay with me as we're in the process of migrating away from snapshots.

src/address-search/data-providers/AddressSearch.js
yeahfro: Can we copy the tests in this PR?

src/address-search/data-providers/AddressSearch.js
paulgain: I tried that and React was complaining about duplicate keys.

src/address-search/__tests__/AddressSearch.test.jsx
paulgain: 👍

src/address-search/__stories__/AddressSearch.stories.jsx
rafenden: Well, `YOUR_GETADDRESS_IO_API_KEY` is more relevant as it tells user what kind of key it is and where it could be obtained, `POSTCODE_KEY` doesn't specify the provider name.

src/address-search/data-providers/AddressSearch.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable camelcase */
```
rafenden: Could we please call this file `PostcodeLookup.js` or similar so it's not confused by the main component name (`AddressSearch.jsx`)?

paulgain: Makes sense, I'll do that in the next PR if that's OK.

PR #156, opened 2019-09-13
feat: EntitySearch refactoring and country support

rafenden:
https://trello.com/c/cHKiVprg/267-support-country-codes-in-entitysearch-component

Changes in this PR:
* updated the way country is presented on the `FieldDnbCompany` field
* added a query parameter to the D&B search to allow for country-based filtering
* replaced "EntitySearchWithDataProvider" with hooks for greater flexibility

## Before
image

## After
image


src/entity-search/EntitySearch.jsx
```diff
@@ -7,14 +7,17 @@ import { SPACING } from '@govuk-react/constants'
...
 import EntityFilters from './EntityFilters'
-import PreviouslySelected from './PreviouslySelected'
 import useFilter from './useFilter'
 
 const StyledButton = styled(Button)`
   margin: ${SPACING.SCALE_2} 0;
 `
 StyledButton.displayName = 'Search'
 
+const StyledPreviouslySelected = styled('div')`
```
mcbhenwood: Small point but really cosmetic changes ought to go into a separate (`style:` ?) commit.

src/entity-search/__mocks__/company-search.js
```diff
@@ -5,32 +5,32 @@ import fixtures from '../__fixtures__'
 
 const mock = new MockAdapter(axios)
 
-export function setupSuccessMocks(apiEndpoint) {
+export function setupSuccessMocks(apiEndpoint, queryParams = {}) {
   mock
-    .onPost(apiEndpoint, {})
+    .onPost(apiEndpoint, { ...queryParams })
```
yeahfro: Why do we need to spread `queryParams` here? It's not as if `onPost` is mutating the data.

rafenden: Good point, this is fixed now!

rafenden:
> Some general feedback on the PR:
> 
> * the change is quite large for a single commit - it really needs to be broken into multiple commits - **DONE**
> * the screenshots do not match the change - **Screenshots updated**
> * the description does not match the change - **Updated**
> 
> I also noticed that the `EntitySearch` stories have not been updated to include the filters.

**EntitySearch** no longer has filters, these are handled in **FieldDnbSearch**.

@yeahfro reply in the quote above

src/entity-search/__mocks__/company-search.js
```diff
@@ -3,34 +3,30 @@ import axios from 'axios'
...
-export function setupSuccessMocks(apiEndpoint) {
+export function setupSuccessMocks(apiEndpoint, adapterOptions = {}, queryParams = {}) {
+  const mock = new MockAdapter(axios, adapterOptions)
   mock
-    .onPost(apiEndpoint, {})
+    .onPost(apiEndpoint, queryParams)
     .reply(200, fixtures.companySearch)
   mock
-    .onPost(apiEndpoint, { search_term: 'some other company' })
+    .onPost(apiEndpoint, { search_term: 'some other company', ...queryParams })
```
yeahfro: There's two more of these `queryParams` using spread

src/entity-search/__mocks__/company-search.js
```diff
@@ -3,34 +3,30 @@ import axios from 'axios'
...
     .reply(200, fixtures.companySearchFilteredByCompanyName)
   mock
-    .onPost(apiEndpoint, { postal_code: 'BN1 4SE' })
+    .onPost(apiEndpoint, { postal_code: 'BN1 4SE', ...queryParams })
```
yeahfro: Here

src/entity-search/__tests__/__snapshots__/useDnbSearch.test.jsx.snap
```diff
@@ -0,0 +1,258 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
yeahfro: Why are we introducing this?

src/entity-search/__mocks__/company-search.js
rafenden: This one needs spreading.

src/entity-search/__mocks__/company-search.js
rafenden: This one needs spreading.

src/entity-search/__tests__/__snapshots__/useDnbSearch.test.jsx.snap
rafenden: To test against the entire response, there is also transformation involved, otherwise I would have to keep the response as a fixture which wouldn't differ much from the snapshot itself.

PR #157, opened 2019-09-13
fix: Correct mixed govuk-react versions

reupen:
We had mixed versions of govuk-react in `package.json` and `yarn.lock`.

This caused both 0.7.0 and 0.7.1 to be specified in `yarn.lock`. This resulted in version 0.7.0 being used in some cases when version 0.7.1 was specified.

This corrects the mixed versions by removing the dependency on `govuk-react` and only using the namespaced `@govuk-react/*` modules.

Note: Only a couple of the `@govuk-react/*` modules are on version 0.7.2.


reupen:
@rafenden @yeahfro Anything worrying in the activity feed snapshot changes?

PR #162, opened 2019-09-17
feat: Show alert when leaving unsaved form

rafenden:
Prevent users from accidental loss of data when reloading form which wasn't submitted.

https://trello.com/c/00net3J8/278-show-alert-when-user-leaves-the-form-which-is-in-progress

## Screenshot

image

### DIFF
--- a/src/forms/hooks/__tests__/useForm.test.jsx
+++ b/src/forms/hooks/__tests__/useForm.test.jsx
@@ -15,6 +15,46 @@ const testField2 = {
 }
 
 describe('useForm', () => {
+  describe('when the hook is called', () => {
+    let hook
+
+    beforeAll(() => {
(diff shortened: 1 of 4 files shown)


paulgain:
@rafenden I'm in this branch and loaded Storybook `(Form - Full example)`. I've refreshed the browser which simply reloads Storybook as expected. The question is... how did you get the Alert to display?

src/forms/hooks/__tests__/useForm.test.jsx
```diff
@@ -445,4 +489,33 @@ describe('useForm', () => {
       expect(error).toBeUndefined()
     })
   })
+
+  describe('when setIsSubmitted() is called', () => {
```
yeahfro: This description does not match what is happening

PR #163, opened 2019-09-17
style: Update spacing of StatusMessage component

rafenden:
This PR adds additional margin-bottom to the `StatusMessage` component.

### DIFF
--- a/src/status-message/StatusMessage.jsx
+++ b/src/status-message/StatusMessage.jsx
@@ -1,16 +1,18 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { SPACING } from '@govuk-react/constants'
+import { spacing } from '@govuk-react/lib'
 import { BLUE } from 'govuk-colours'
+import { FOCUSABLE, SPACING } from '@govuk-react/constants'
 
 const StyledStatusMessage = styled('div')`
   border: ${({ colour }) => `${SPACING.SCALE_1} solid ${colour}`};
   color: ${({ colour }) => colour};
-  padding: ${SPACING.SCALE_3};
-  margin-top: ${SPACING.SCALE_2};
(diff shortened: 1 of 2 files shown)


src/status-message/StatusMessage.jsx
```diff
@@ -1,21 +1,23 @@
...
   line-height: 1.5;
+  ${spacing.responsivePadding(4)};
+  ${spacing.withWhiteSpace({ marginBottom: 6 })};
+  ${FOCUSABLE};
 `
 
 const StatusMessage = ({ colour, children }) => {
   return (
-    <StyledStatusMessage colour={colour}>
+    <StyledStatusMessage tabIndex={0} colour={colour}>
```
ian-leggett: Do we need a tab index here?

rafenden: tabindex removed, thanks!

src/status-message/StatusMessage.jsx
```diff
@@ -1,16 +1,18 @@
...
 const StyledStatusMessage = styled('div')`
   border: ${({ colour }) => `${SPACING.SCALE_1} solid ${colour}`};
   color: ${({ colour }) => colour};
-  padding: ${SPACING.SCALE_3};
-  margin-top: ${SPACING.SCALE_2};
   font-weight: bold;
   line-height: 1.5;
+  ${spacing.responsivePadding(4)};
+  ${spacing.withWhiteSpace({ marginBottom: 6 })};
+  ${FOCUSABLE};
```
yeahfro: This is missing from test

rafenden: I'm not convinced whether adding tests for each CSS rule is a good idea.

rafenden: I added the test.

PR #164, opened 2019-09-17
fix: Make EntityListItem focusable

rafenden:
This change will let users tab through entity search results.

### DIFF
--- a/src/entity-search/EntityListItem.jsx
+++ b/src/entity-search/EntityListItem.jsx
@@ -1,17 +1,20 @@
 import React from 'react'
 import styled from 'styled-components'
-import { FONT_SIZE, MEDIA_QUERIES, SPACING } from '@govuk-react/constants'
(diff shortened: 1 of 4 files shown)


src/entity-search/EntityListItem.jsx
```diff
@@ -47,7 +48,9 @@ const EntityListItem = ({ canHandleClick, onEntityClick, data, text, heading, me
   return (
     <StyledEntity
       key={uniqueId()}
-      onClick={() => (canHandleClick ? onEntityClick(data) : null)}
+      tabIndex={canHandleClick ? 0 : false}
+      onClick={() => canHandleClick && onEntityClick(data)}
+      onKeyDown={e => canHandleClick && e.keyCode === 13 && onEntityClick(data)}
```
yeahfro: This could be tested

rafenden: Thanks, tests added!

PR #166, opened 2019-09-18
fix: Set limit on the company name filter in FieldDnbCompany

rafenden:
https://trello.com/c/WxE5JdB3/334-character-limit-on-company-name-search-field-should-be-enforced

Required due to limitation of the DnB service.

### DIFF
--- a/src/forms/hooks/__tests__/useForm.test.jsx
+++ b/src/forms/hooks/__tests__/useForm.test.jsx
@@ -508,23 +508,6 @@ describe('useForm', () => {
     })
   })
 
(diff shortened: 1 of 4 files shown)


src/forms/hooks/__tests__/useForm.test.jsx
```diff
@@ -490,23 +490,6 @@ describe('useForm', () => {
     })
   })
 
-  describe('when setIsSubmitted() is called', () => {
```
rafenden: It's a duplicated test added by mistake in previous PR.

PR #167, opened 2019-09-19
feat: Add address lookup UI components

paulgain:
## Test within Storybook
Forms/FieldAddress and requires an API token as part of `knobs`.

## UK
Sep-19-2019 13-12-03

## FOREIGN
foreign

## NOTE:
Tests to follow (working on them now) allowing @yeahfro time to integrate.


src/address-search/AddressSearch.jsx
```diff
@@ -3,43 +3,70 @@ import PropTypes from 'prop-types'
...
+
   return (
     <>
-      <div>
+      <FieldWrapper name="postcode" label="Postcode">
         <InputField
-          name="postcode"
-          type="text"
           error={error}
+          style={{ width: '165px', marginBottom: '30px' }}
```
yeahfro: Can we swap out these styles in favour of `styled-components`?

src/address-search/AddressSearch.jsx
```diff
@@ -3,43 +3,70 @@ import PropTypes from 'prop-types'
...
-              return <option key={value.id} value={index}>{ value.address1 }</option>
-            })
-          }
-        </Select>
+        <FieldWrapper name="address" label="Select an address">
+          <Select
+            value={address}
+            onChange={onAddressChange}
+          >
+            { addressList.map((obj, index) => {
```
yeahfro: Could be nice to destructure `obj`

src/address-search/AddressSearch.jsx
```diff
@@ -3,43 +3,70 @@ import PropTypes from 'prop-types'
...
+              return <option key={obj.id} value={index}>{ obj.address1 }</option>
```
yeahfro: Are the extra spaces required either side of `obj.address1`?

src/address-search/data-providers/AddressSearch.js
```diff
@@ -35,9 +35,20 @@ const transformAddress = (address, postcode) => {
   }
 }
 
+const addAddressCount = (addresses) => {
+  const address = addresses.length === 1 ? 'address' : 'addresses'
```
yeahfro: Using `pluralise` would work well here.

src/address-search/data-providers/AddressSearch.js
```diff
@@ -35,9 +35,20 @@ const transformAddress = (address, postcode) => {
...
+  })
+
+  return addresses
+}
+
 export default (baseUrl, apiKey) => {
   return async (postcode) => {
     const { data } = await axios.get(`${baseUrl}/${postcode}?api-key=${apiKey}`)
-    return map(data.Addresses, address => transformAddress(address, postcode))
+    return addAddressCount(map(data.Addresses, address => transformAddress(address, postcode)))
```
yeahfro: Breaking this over multiple lines would make it easier to read.

In addition, rather than mutating your array (and maintaining immutability), you could do:

```
export default (baseUrl, apiKey) => {
  return async (postcode) => {
    const { data } = await axios.get(`${baseUrl}/${postcode}?api-key=${apiKey}`)
    const mapped = map(data.Addresses, address => transformAddress(address, postcode));

    return [
      {
        // new obj
      },
      ...mapped,
    ]
  }
}
```

src/forms/elements/FieldAddress.jsx
```diff
@@ -0,0 +1,69 @@
...
+  legend,
+  apiKey,
+  apiEndpoint,
+  label,
+  hint,
+}) => {
+  const { setFieldValue } = useFormContext()
+  const isUK = country === UNITED_KINGDOM
+
+  function onAddressSelect(address) {
```
yeahfro: What could be quite nice is if these match what's in the API, ie:

```
    address: {
      line_1: 'Unit 10, Ockham Drive',
      line_2: null,
      town: 'GREENFORD',
      county: null,
      country: {
        id: '80756b9a-5d95-e211-a939-e4115bead28a',
      },
      postcode: 'UB6 0F2',
    },
```

Better to avoid the transformation if we can.

src/address-search/AddressSearch.jsx
paulgain: 👍

src/address-search/AddressSearch.jsx
paulgain: 👍

src/address-search/AddressSearch.jsx
paulgain: Nope, removed.

src/address-search/data-providers/AddressSearch.js
paulgain: 👍

src/forms/elements/FieldAddress.jsx
paulgain: This is the response from the FE (e.g. Node) postcode lookup **which my changes now match:**
http://localhost:3000/api/postcodelookup/SW1H9AJ
```
{
  address1: "Ministry of Justice",
  address2: " 102 Petty France",
  county: " ",
  city: " London",
  postcode: "SW1H 9AJ",
  country: "80756b9a-5d95-e211-a939-e4115bead28a"
},
```

yeahfro: My thinking here was that the shape of the `address` `object` above is what is expected by Leeloo. Currently there is some extra work that needs to be done to transform the request body to the correct shape. Maybe we can follow up with another PR to simplify this and be okay with it for now.

paulgain: Follow up PR it is.

PR #171, opened 2019-09-20
fix: Replace InputField with Input component to remove extra label tag

rafenden:
This PR is removing the additional `label` tag added by`InputField` component.

No visual changes.

### DIFF
--- a/src/forms/elements/FieldInput.jsx
+++ b/src/forms/elements/FieldInput.jsx
@@ -1,28 +1,42 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import InputField from '@govuk-react/input-field'
+import styled from 'styled-components'
+
+import Input from '@govuk-react/input'
+import ErrorText from '@govuk-react/error-text'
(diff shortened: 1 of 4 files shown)


paulgain:
Note: this removes not only an additional label but also the child `span`.

src/forms/elements/FieldInput.jsx
```diff
@@ -1,28 +1,42 @@
...
+import Input from '@govuk-react/input'
+import ErrorText from '@govuk-react/error-text'
+import { BORDER_WIDTH_FORM_ELEMENT_ERROR, SPACING } from '@govuk-react/constants'
+import { ERROR_COLOUR } from 'govuk-colours'
 
 import useField from '../hooks/useField'
 import FieldWrapper from './FieldWrapper'
 
+const StyledInputWrapper = styled('div')`
+  ${props => props.error && `
```
yeahfro: If I comment out all of this then the tests still pass. The reason for caring about this in particular is that there is a condition in the style.

rafenden: Good point, test added!

PR #177, opened 2019-09-24
fix: Set country as country ID

yeahfro:
## Description
Sets the country field value as the country ID.


src/forms/elements/FieldAddress.jsx
```diff
@@ -23,15 +23,15 @@ const FieldAddress = ({
   } = useAddressSearch(postcodeLookup(apiEndpoint))
 
   const { setFieldValue } = useFormContext()
-  const isUK = country === UNITED_KINGDOM
+  const isUK = country.name === UNITED_KINGDOM
```
rafenden: Shall we use the ISO code here?

PR #183, opened 2019-09-26
feat: Allow for custom components as back and forward buttons in forms

rafenden:
**BREAKING CHANGE:** This change is replacing `hideForwardButton`, `hideBackButton`, `forwardButtonText` and `backButtonText` with new `backButton` and `forwardButton` props which now accepts React elements.

**API changes**

`<Step hideForwardButton={true} />` **becomes** `<Step forwardButton={null} />`
`<Step forwardButtonText="Submit form" />` **becomes** `<Step forwardButton="Submit form" />`

New one:
`<Step backButton={<Link href="/companies">Back</Link>} />`

### DIFF
--- a/src/forms/elements/__tests__/Step.test.jsx
+++ b/src/forms/elements/__tests__/Step.test.jsx
@@ -16,7 +16,7 @@ describe('Step', () => {
   let formState
   let wrapper
 
-  describe('when there is a form with one step', () => {
+  describe('when there is a form with only one step', () => {
     const onSubmitSpy = jest.fn()
 
     beforeAll(() => {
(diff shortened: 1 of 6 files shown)


src/forms/elements/__tests__/Step.test.jsx
```diff
@@ -321,12 +303,12 @@ describe('Step', () => {
...
+  describe('when the "backButton" prop is passed as "null"', () => {
     beforeAll(() => {
       wrapper = mount(
         <Form initialStep={1}>
           <Step name="testStep1" />
-          <Step name="testStep2" hideBackButton={true} />
+          <Step name="testStep2" backButton={null} />
         </Form>,
       )
     })
```
mcbhenwood: I know it's not part of this PR, but I'm not very fond of the test which validates the hiding of the back button which follows on here at [line 316](https://github.com/uktrade/data-hub-components/blob/a486d16ac43cf848b3679196006d4f0e7ea1c43e/src/forms/elements/__tests__/Step.test.jsx#L316):

```javascript
    test('should hide the back button', () => {
      expect(wrapper.find('button').text().includes('Back')).toBeFalsy()
    })
```
Checking that it doesn't show "Back" seems to miss the point a little. Also this test would be satisfied if you passed in `backButton="Go to previous step"` so I don't feel it's strong enough. Maybe it could be improved to test the visibility of the actual button? [The corresponding test for the 'forward' button uses `.exists()` which seems much stronger.]

rafenden: Good point, I added a name to the button so the test looks like that now:
```
expect(wrapper.find('button[name="back"]').exists()).toBeFalsy()
```

PR #188, opened 2019-09-30
chore: add new eslint rules for empty lines, indents and EOF line breaks

debitan:
**To go in ahead of #189**

To avoid having to check for these formatting issues in the PR stage I've added new eslint rules to restrict empty line to a max of 1, indents to 2 characters and to force a line break at the end of files. 

I've also run `--fix` to remove the errors currently in the codebase. 

This might not all be desirable so is totally open to suggestions.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,40 +19,86 @@
     "react-hooks"
   ],
   "rules": {
-    "no-underscore-dangle": ["error", {"allow": ["_source"]}],
+    "no-underscore-dangle": [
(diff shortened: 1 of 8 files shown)


debitan:
> It would have made sense if this was two commits:
> 
> 1. Update code format
> 2. Add ESLint rules

can do that, no problem. are you happy with the eslint rules?

debitan:
@yeahfro now broken into two commits

PR #192, opened 2019-10-02
feat: Add CollectionDownload component

debitan:
## Description of change

Add `CollectionDownload` component.

The component displays four different things, depending on the number of items passed to it. 

This is the third part of a large piece of work to build out the CollectionList template in React in order to show a list of Large Capital Profiles (and possibly in the future to replace the CollectionList nunjucks macro in the frontend).

## Screenshot

With no items:
no items

With 1 item:
1 item

With 2 - 5000 items:
101 items

With 5001+ items:
5001 items

The aim is to get to something resembling the below screenshot:

### DIFF
--- a/src/collection-list/CollectionDownload.jsx
+++ b/src/collection-list/CollectionDownload.jsx
@@ -1,0 +1,94 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import pluralise from 'pluralise'
+import {
+  SPACING,
+  MEDIA_QUERIES,
+  BORDER_WIDTH_FORM_ELEMENT,
+} from '@govuk-react/constants'
+import Button from '@govuk-react/button'
+import Paragraph from '@govuk-react/paragraph'
+import { BLACK } from 'govuk-colours'
+
+const StyledWrapper = styled('div')`
+  display: flex;
+  flex-flow: row wrap;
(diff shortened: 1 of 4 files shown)


src/collection-list/CollectionDownload.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const PluralText = `You can download these ${itemPluralWithCount}`
+
+  const NoItemsText = `There are no ${itemPlural} to download`
+  const DownloadText = (totalItems > 1 ? PluralText : SingularText)
+  const NeedToFilterText = `Filter to fewer than 5000 ${itemPlural} to download`
+
+  if (totalItems === 0) {
+    return (
+      <StyledWrapper>
+        <StyledParagraph>{NoItemsText}</StyledParagraph>
```
rafenden: I think it's fine to display the test directly here.

debitan:
> Looks good. One thing I noticed which is very small is the jump between have a collection to download to not having one to download. Example
> Screenshot 2019-10-07 at 10 26 42
> Screenshot 2019-10-07 at 10 26 35

@ian-leggett 
The only way I can see round this is to add `min-height: 49px` (which is the height of the component when the download button is showing).

However, this would only fix it for desktop. On mobile the download button appears below and at full width, so if we fix the height there too there would be a load of whitespace. 

Also, it requires using `px`. There is no constant in the govuk-react library that fits the bill for this. So my feeling is to leave it as it is?

PR #197, opened 2019-10-08
feat: add CollectionPagination component

debitan:
## Description of change

Add `CollectionPagination` component.

The component displays only the previous and next button on mobile view (if passed) and does not render at all if there is only 1 page. (the below screenshot is an exception as I've passed 7 pages to the component in storybook - despite it saying 1 profile).

This is the fourth part of a large piece of work to build out the CollectionList template in React in order to show a list of Large Capital Profiles (and possibly in the future to replace the CollectionList nunjucks macro in the frontend).

## Screenshot

Desktop view:
Screenshot 2019-10-08 at 13 51 05

Mobile view:
Screenshot 2019-10-08 at 13 51 53

### DIFF
--- a/src/collection-list/CollectionPagination.jsx
+++ b/src/collection-list/CollectionPagination.jsx
@@ -80,7 +80,7 @@ function CollectionPagination({
   next,
   pages,
 }) {
-  function getPages(page, current) {
+  function getPage(page, current) {
     if (!page.url) {
       return <StyledSpan>{page.label}</StyledSpan>
     }
@@ -95,7 +95,7 @@ function CollectionPagination({
       {previous && <StyledPrevious href={previous}>Previous</StyledPrevious>}
       <StyledList>
         {pages.map((page) => (
-          <StyledListItem>{getPages(page, currentPage)}</StyledListItem>
+          <StyledListItem>{getPage(page, currentPage)}</StyledListItem>
         ))}
       </StyledList>
       {next && <StyledNext href={next}>Next</StyledNext>}
--- a/src/collection-list/CollectionPagination.jsx
+++ b/src/collection-list/CollectionPagination.jsx
@@ -84,7 +84,7 @@ function CollectionPagination({
     if (!page.url) {
       return <StyledSpan>{page.label}</StyledSpan>
     }
-    if (page.label === JSON.stringify(current)) {
+    if (page.label === String(current)) {
       return <StyledCurrentAnchor>{page.label}</StyledCurrentAnchor>
     }
     return <StyledAnchor href={page.url}>{page.label}</StyledAnchor>
--- a/src/collection-list/CollectionPagination.jsx
+++ b/src/collection-list/CollectionPagination.jsx
@@ -1,0 +1,127 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { SPACING, MEDIA_QUERIES, FONT_SIZE } from '@govuk-react/constants'
+import { GREY_4, LINK_COLOUR, GREY_3, TEXT_COLOUR, GREY_1 } from 'govuk-colours'
+
+const StyledNav = styled('nav')`
+  text-align: center;
+  line-height: 1;
+  display: flex;
+  justify-content: space-around;
+  padding: ${SPACING.SCALE_3};
+
+  ${MEDIA_QUERIES.TABLET} {
+    display: block;
+  }
+`
+
+const StyledAnchor = styled('a')`
+  font-weight: bold;
+  font-size: ${FONT_SIZE.SIZE_16};
+  display: inline-block;
+  padding: ${SPACING.SCALE_2} ${SPACING.SCALE_3};
+  background-color: ${GREY_4};
+  line-height: ${FONT_SIZE.SIZE_24};
+  color: ${LINK_COLOUR};
+  text-decoration: none;
+
+    :hover {
+      background-color: ${GREY_3};
+    }
+  }
+`
+
+const StyledCurrentAnchor = styled(StyledAnchor)`
+  background-color: transparent;
+  color: ${TEXT_COLOUR};
+  text-decoration: none;
+`
+
+const StyledPrevious = styled(StyledAnchor)`
+  margin-right: ${SPACING.SCALE_1};
+`
+
+const StyledNext = styled(StyledAnchor)`
+  margin-left: ${SPACING.SCALE_1};
+`
+
+const StyledList = styled('ul')`
+  display: none;
+
+  ${MEDIA_QUERIES.TABLET} {
+    display: inline-block;
+    padding: 0;
+  }
+`
+
+const StyledListItem = styled('li')`
+  display: inline-block;
+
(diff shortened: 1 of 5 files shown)


src/collection-list/CollectionPagination.jsx
```diff
@@ -0,0 +1,127 @@
...
+              </StyledListItem>
+            )
+          } else {
+            return (
+              <StyledListItem>
+                <StyledSpan>{page.label}</StyledSpan>
+              </StyledListItem>
+            )
+          }
+        })}
```
web-bert: I think the repeated `StyledListItem` is not very DRY, might be better to split that logic out into a function:
```
function getItemContent(page, currentPage) {
   if (page.url) {
      if (page.label === JSON.stringify(currentPage)) {
         return (
            <StyledCurrentAnchor>{page.label}</StyledCurrentAnchor>
         )
      } else {
         return (
            <StyledAnchor href={page.url}>{page.label}</StyledAnchor>
         )
      }   
   } else {
      return (
            <StyledSpan>{page.label}</StyledSpan>
      )
   }
}
<StyledList>
   {pages.map((page) => (
      <StyledListItem>
        {getItemContent(page, currentPage)}
      </StyledListItem>
   ))}
</StyledList>
```

src/collection-list/CollectionPagination.jsx
```diff
@@ -0,0 +1,127 @@
...
+  color: ${GREY_1};
+`
+function CollectionPagination({
+  totalPages,
+  currentPage,
+  previous,
+  next,
+  pages,
+}) {
+  return totalPages === 1 ? null : (
```
yozzo: there's a lot of logic here, could we break it down into separate functions, also refactor the multi line ternary operator to an if/else statement?

src/collection-list/CollectionPagination.jsx
```diff
@@ -0,0 +1,127 @@
...
+            )
+          }
+        })}
+      </StyledList>
+      {next && <StyledNext href={next}>Next</StyledNext>}
+    </StyledNav>
+  )
+}
+
+CollectionPagination.propTypes = {
```
yozzo: `propTypes` and `defaultProps` usually sit at the top of the component

debitan: in class based components they do. but for functional components they tend to sit at the bottom.

debitan:
> How do we re-use the logic to decide how many pages to show, how many there are etc. Should this have `totalPages`, `getPageUrl` and `currentPage` props where all the logic can be inside this component? i.e. it decides when to show the `...` or not based on `totalPages` and `currentPage` and then calls `getPageUrl` with a `pageIndex` to get the url?

@web-bert looking at the logic on the frontend it looks like the `...` is passed into the pagination nunjucks component (the current implementation), so I think that logic is handled in the frontend. fully prepared to refactor if that isn't the case though and include that logic in here.

I haven't pulled this through to the frontend to play with yet (partly because of the problems we discussed in the huddle this morning about doing that before the package is published), so that is why I haven't fully explored this yet.

web-bert:
@debitan part of the reason the `...` and other logic is not part of the current macro is because macros are not really designed to have complex logic. Now that we have the convenience of JavaScript to use in the component it seems to make sense to have that logic as part of the component. Otherwise it would mean there is a prerequisite of running the data through something to get a list of pages that are then passed to this, which ultimately makes this less useful and is asking for inconsistencies to start cropping up. Happy for others to refute this though...

debitan:
@web-bert I think that is a fair point. 

I guess the question is whether it is something for this PR to tackle or part of a future refactor of the component when it is time to implement the CollectionList in the frontend?

src/collection-list/__stories__/CollectionList.stories.jsx
```diff
@@ -30,6 +32,13 @@ storiesOf('Collection', module).add('Collection', () => (
       badges={capitalProfileItem.badges}
       metadata={capitalProfileItem.metadata}
     />
+    <CollectionPagination
+      totalPages={paginationProps.totalPages}
```
rafenden: We don't need to pass `totalPages` as this can be calucated from `page`.

src/collection-list/CollectionPagination.jsx
```diff
@@ -72,38 +72,31 @@ const StyledSpan = styled('span')`
...
-            return (
-              <StyledListItem>
-                <StyledSpan>{page.label}</StyledSpan>
-              </StyledListItem>
-            )
-          }
-        })}
+        {pages.map((page) => (
+          <StyledListItem>{getPages(page, currentPage)}</StyledListItem>
+        ))}
```
yozzo: having loops in the template or any kind of logic make is _harder_ to understand later on, due to the cornucopia of variables, custom tags, curly brackets, brackets, arrows etc. I understand that at the moment the logic is recent in the developer's mind, but after a few sprints when this component needs revisiting, whoever picks it up will need to decipher it first, and then enhance it.  Caching the logic inside a function outside the render will dilute it a bit:
```javascript
function renderPages(pages) {
return (pages.map((page) => (
     <StyledListItem>{getPages(page, currentPage)}</StyledListItem>
))
//...
<StyledList>
  {this.renderPages(pages)}
</StyledList>
```

src/collection-list/__stories__/CollectionList.stories.jsx
debitan: At the moment it can't as there is some logic in the frontend that takes the pages and shows a max of 6 pages (including an added `...` truncation page). So the total pages could be 102, but the max number of pages would be 6.

@web-bert has a good point about this logic sitting in the component now that we have access to JavaScript, but I'm unsure whether I should add this at this stage and think that it should probably be a separate piece of work.

src/collection-list/CollectionPagination.jsx
```diff
@@ -0,0 +1,120 @@
...
+`
+
+function CollectionPagination({
+  totalPages,
+  currentPage,
+  previous,
+  next,
+  pages,
+}) {
+  function getPages(page, current) {
```
web-bert: This should be `getPage` as it only returns one

src/collection-list/CollectionPagination.jsx
```diff
@@ -0,0 +1,120 @@
...
+  currentPage,
+  previous,
+  next,
+  pages,
+}) {
+  function getPage(page, current) {
+    if (!page.url) {
+      return <StyledSpan>{page.label}</StyledSpan>
+    }
+    if (page.label === JSON.stringify(current)) {
```
web-bert: What is the need for JSON.stringify here?

debitan: @web-bert because of our linter. I can't use a straight up `===` because the two things aren't the same type. So I started with a `==`, but then the linter shouted at me.

web-bert: I assume current is a Number and label is a string so `if (page.label === String(current))` would be better

debitan: that would also work. is that preferable?

PR #198, opened 2019-10-08
feat: filter activity feed by type of activity

yozzo:
Implementation of activity type filter for activity feed. With the use of a dropdown, a user can select which activities to see in the feed. For this iteration we have *All Activities* *Data Hub Activities*, and *External Activities*.

The previous filter which only expands the details of each activity still exists and it is enabled automatically when no constant values attributed to the activity type filter.

Screenshot 2019-10-14 at 09 24 30


yozzo:
> I can see this is a `WIP` from the tag and you'll already know you'll need to do this, but as you've asked for reviews (and it'll bump my Pull Panda stats) 😈 :
> 
> * this needs to be broken down into smaller PRs
> * it needs a description and preferably some screenshots
> * the test coverage needs improving and transitioning away from snapshots

All three PR's are under 500 lines

src/activity-feed/constants.js
```diff
@@ -2,4 +2,41 @@ export const SOURCE_TYPES = {
...
+]
+
+export const ACTIVITY_TYPE = {
+  CompaniesHouseAccount: ['dit:Accounts'],
+  CompaniesHouseCompany: ['dit:Company'],
+  HmrcExporter: ['dit:Export'],
+  Interaction: ['dit:Interaction', 'dit:ServiceDelivery'],
+  InvestmentProject: ['dit:InvestmentProject'],
+  Omis: ['dit:OMISOrder'],
+}
```
web-bert: It might be a good idea to have an `ACTIVITY_FEED_TYPES` object with all the `'dit:Company'`, 'dit:Interaction' etc there or re-use the `ACTIVITY_TYPE` options in the `ACTIVITY_TYPE_FILTERS` to make this a bit more DRY and reduce any potential typos

src/activity-feed/constants.js
```diff
@@ -2,4 +2,44 @@ export const SOURCE_TYPES = {
...
 }
 
-export default { SOURCE_TYPES }
+export const ACTIVITY_TYPE = {
+  CompaniesHouseAccount: ['dit:Accounts'],
+  CompaniesHouseCompany: ['dit:Company'],
+  HmrcExporter: ['dit:Export'],
+  Interaction: ['dit:Interaction', 'dit:ServiceDelivery'],
+  InvestmentProject: ['dit:InvestmentProject'],
+  Omis: ['dit:OMISOrder'],
```
web-bert: These should really all start lowercase, but not a big issue

src/activity-feed/constants.js
```diff
@@ -2,4 +2,44 @@ export const SOURCE_TYPES = {
...
+}
+
+export const ACTIVITY_TYPE_FILTERS = [
+  {
+    label: 'All Data Hub & external activity',
```
web-bert: Wouldn't this be better as just "All activity"

src/activity-feed/constants.js
yozzo: this is how they're coming from the API, there's some naming convention [https://www.w3.org/TR/activitystreams-vocabulary/#termconventions](url) that requires types to be capitalised, but might clash with the prefixing TL;DR -> all Activity Stream APIs follow the same convention

src/activity-feed/constants.js
yozzo: I had the same thought but there's also a filter called _My Activity_ that might create confusion.

src/activity-feed/ActivityFeed.jsx
```diff
@@ -48,12 +49,24 @@ export default class ActivityFeed extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
+      filteredActivity: [],
       showDetails: false,
+      isActivityTypeFilterEnabled: !!ACTIVITY_TYPE_FILTERS,
```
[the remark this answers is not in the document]
yozzo: good call. I amended it to be false when the `ACTIVITY_TYPE_FILTERS` is an empty list.

src/activity-feed/constants.js
rafenden: Agree with @web-bert, we could also rename `My activity` => `Only my activity`, would be good to talk to Rogier about this.

yozzo: I followed their design

src/activity-feed/filters/BasicActivityTypeFilter.jsx
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  input {
+    width: 26px;
+    height: 26px;
+    margin: 0;
+    padding: 0;
+  }
+`
```
debitan: Might be better to use the constants here, e.g. `border-width: 0 0 ${BORDER_WIDTH_MOBILE} ${BORDER_WIDTH_MOBILE}`

src/activity-feed/ActivityFeed.jsx
```diff
@@ -48,12 +49,24 @@ export default class ActivityFeed extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
+      filteredActivity: ACTIVITY_TYPE_FILTERS[2].value,
       showDetails: false,
+      isActivityTypeFilterEnabled: !!ACTIVITY_TYPE_FILTERS.length,
```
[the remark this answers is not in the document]
yozzo: it means it will always return a boolean, even though any integer different than 0 is truthy, for semantic reasons I'm making it a boolean

src/activity-feed/filters/BasicActivityTypeFilter.jsx
```diff
@@ -0,0 +1,105 @@
...
+      left: 5px;
+      width: 12px;
+      height: 6px;
+      border-width: 0 0 ${BORDER_WIDTH_MOBILE} ${BORDER_WIDTH_MOBILE};
+    }
+  }
+
+  input {
+    width: 26px;
+    height: 26px;
```
debitan: can we use the constants instead of all the `px` sizes and `white`?

yozzo: updated the white with a constant, but the `px`'s will need to stay as they are because we don't have corresponding variables for `26px` or `6px` unless I start calculating them from e.g. `${ (GUTTER_PX / 6) + 1 }` and I don't think that's the expected use.

Replacing with closest var values i.e. `5px`, `25px` etc will render unwanted results as you can see below:

original:
Screenshot 2019-10-18 at 18 27 41

optimised to using closest dimensions provided by variables:
Screenshot 2019-10-18 at 18 27 07

debitan: Fair enough. I agree we shouldn't start doing things like `${ (GUTTER_PX / 6) + 1 }`. In my components I deviated from the design a bit to be able to use constants, but it still looked ok anyway.

src/activity-feed/filters/BasicActivityTypeFilter.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+    &::before {
+      background: ${WHITE};
+      width: 26px;
+      height: 26px;
+    }
+
+    &::after {
+      top: 6px;
+      left: 5px;
```
debitan: this one could be `${SPACING.SCALE_1}`

src/activity-feed/filters/BasicActivityTypeFilter.jsx
```diff
@@ -0,0 +1,105 @@
...
+      width: 12px;
+      height: 6px;
+      border-width: 0 0 ${BORDER_WIDTH_MOBILE} ${BORDER_WIDTH_MOBILE};
+    }
+  }
```
debitan: Actually, I think it would be preferable if this span was it's own styled component, rather than it's styling sitting within that of the checkbox. It makes it cleaner and clearer which styles affect which tag. But I don't know if it would mess up the styling or not. 

I'm guessing the `input` styling below is harder to break out as it is a inherent part of the `Checkbox` govuk react component. But the span should be able to be wrapped with a styled component.

src/activity-feed/filters/SelectFilter.jsx
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+    
+    select {
+      width: 100%;
+      
+      ${MEDIA_QUERIES.DESKTOP} {
+        width: auto;
+      }
+    }
```
debitan: Same thing here. The `label`, `span` and `select` should have their own styled components rather than being nested within the `StyledDropdownContainer`.

yozzo: `span` and `label` selectors are used to overwrite the properties encapsulated in the govUK `Select` component. 

Breaking it down in styled components means I'll need to create an extra wrapping element and then pass in the properties we're overriding.

src/activity-feed/filters/BasicActivityTypeFilter.jsx
yozzo: the `& > span` selector is referring to an encapsulated element inside the govUK `Checkbox` component

src/activity-feed/filters/SelectFilter.jsx
debitan: fair enough

src/activity-feed/filters/BasicActivityTypeFilter.jsx
debitan: you could wrap the text in another styled component, but I see your point.

PR #201, opened 2019-10-17
Style to be more inline with the data hub updates section

ian-leggett:
Small style additions to make this more consistent with the section above "Data Hub Updates" section. Header was too large and we were missing the border at the top of the section.

### DIFF
--- a/src/dashboard/my-companies/ListSelector.jsx
+++ b/src/dashboard/my-companies/ListSelector.jsx
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 import { H2, H3 } from '@govuk-react/heading'
 import Link from '@govuk-react/link'
 import Select from '@govuk-react/select'
-import { SPACING } from '@govuk-react/constants'
+import { SPACING, LEVEL_SIZE } from '@govuk-react/constants'
 import useMyCompaniesContext from './useMyCompaniesContext'
 import { LIST_CHANGE } from './constants'
 
@@ -12,7 +12,7 @@ const StyledRoot = styled.div({
   alignItems: 'baseline',
 })
 
-const StyledHedline = styled(H2)({
+const StyledHeadline = styled(H2)({
   flexGrow: 1,
 })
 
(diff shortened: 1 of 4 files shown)


src/dashboard/my-companies/ListSelector.jsx
```diff
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 import { H2, H3 } from '@govuk-react/heading'
 import Link from '@govuk-react/link'
 import Select from '@govuk-react/select'
-import { SPACING } from '@govuk-react/constants'
+import { SPACING, LEVEL_SIZE } from '@govuk-react/constants'
```
debitan: what is `LEVEL_SIZE`? I can't see it in the constants file.

src/dashboard/my-companies/MyCompaniesTile.jsx
```diff
@@ -1,10 +1,18 @@
...
+import { SPACING } from '@govuk-react/constants'
+import { GREY_2 } from 'govuk-colours'
+import styled from 'styled-components'
 
 import useMyCompaniesContext from './useMyCompaniesContext'
 import MyCompaniesTable from './MyCompaniesTable'
 import ListSelector from './ListSelector'
 
+const StyledDiv = styled.div({
+  borderTop: `4px solid ${GREY_2}`,
```
debitan: you could use `BORDER_WIDTH_MOBILE` instead of `4px` here

src/dashboard/my-companies/ListSelector.jsx
```diff
@@ -43,7 +43,7 @@ export default () => {
   } = useMyCompaniesContext()
   return (
     <StyledRoot>
-      <StyledHedline>My companies list</StyledHedline>
+      <StyledHeadline size={LEVEL_SIZE[3]}>My Companies Lists</StyledHeadline>
```
rafenden: Why not to use `H3`? I assume it's because `H2` is used on the other tiles?

ian-leggett: Yeah its just down to semantics, the heading level should be a `h2` as its on the same level as the section above which is "Data hub updates"

src/dashboard/my-companies/ListSelector.jsx
ian-leggett: Yeah bit confusing this one, it comes from 'typography' which is exported from constants so it can be imported at the same level as the likes of `SPACING` 
Screenshot 2019-10-18 at 08 58 37

src/dashboard/my-companies/MyCompaniesTile.jsx
ian-leggett: Good spot 👍

PR #203, opened 2019-10-21
Feature/delete list link

peterhudec:
# Change the label and behavior of the my companies _edit lists_ link

We decided that there's gonna be no _edit lists_ page and moved the functionality of removing a list to the dashboard. So now instead of the _edit list_ link to the right of the _companies list selector_, there will be a _delete this list_ link.

Screen Shot 2019-10-21 at 1 00 00 PM
Screen Shot 2019-10-21 at 12 59 37 PM

### DIFF
--- a/src/dashboard/my-companies/ListSelector.jsx
+++ b/src/dashboard/my-companies/ListSelector.jsx
@@ -37,9 +37,9 @@ const StyledSelect = styled(Select)({
 
 export default () => {
   const {
-    state: { lists },
+    state: { lists, selectedIdx },
     dispatch,
-    editListsLinkProps,
+    deleteListPropsAccessor = () => ({}),
   } = useMyCompaniesContext()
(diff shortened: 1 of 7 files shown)


src/dashboard/my-companies/ListSelector.jsx
```diff
@@ -37,9 +37,9 @@ const StyledSelect = styled(Select)({
 
 export default () => {
   const {
-    state: { lists },
+    state: { lists, selectedIdx },
     dispatch,
-    editListsLinkProps,
+    deleteListPropsAccessor = () => ({}),
```
ian-leggett: Do we need this empty default function here?

peterhudec: The other way would be to check for whether it's defined just before the call. I find it clearer this way.

PR #208, opened 2019-10-23
style: Add Jest linting to ESLint

rafenden:
Add linting for Jest tests using https://github.com/jest-community/eslint-plugin-jest


rafenden:
> This looks like it's doing more than just adding lint rules, should this be two 

This PR also has to apply the new rules to pass the tests.

src/activity-feed/__tests__/ActivityFeedPagination.test.jsx
```diff
@@ -17,13 +17,6 @@ describe('ActivityFeedPagination', () => {
     expect(tree).toMatchSnapshot()
   })
 
-  test('renders pagination with loader', () => {
```
rafenden: Duplicated test.

debitan:
It looks like some tests are deleted entirely, is this expected?

rafenden:
> It looks like some tests are deleted entirely, is this expected?

These tests were duplicated and the linter detected it.

PR #209, opened 2019-10-23
fix: company lists width size and cell alignment

ian-leggett:
## Description
This sets fixed percentage widths on the table cells and removes the centre alignment on the "Last interaction" cells.


paulgain:
LGTM

PR #212, opened 2019-10-24
fix: count only filtered activities

yozzo:
currently it's showing the count for all activities coming from the API, we need to display only the count for the activities in view. (This creates confusion when a large number of activities are present) 
Screenshot 2019-10-24 at 11 44 35

### DIFF
--- a/src/activity-feed/ActivityFeed.jsx
+++ b/src/activity-feed/ActivityFeed.jsx
@@ -1,3 +1,4 @@
+import { intersection } from 'lodash'
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
@@ -32,7 +33,6 @@ export default class ActivityFeed extends React.Component {
     isLoading: PropTypes.bool,
     addContentText: PropTypes.string,
     addContentLink: PropTypes.string,
-    totalActivities: PropTypes.number,
   }
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/activity-feed/ActivityFeed.jsx
```diff
@@ -61,9 +60,10 @@ export default class ActivityFeed extends React.Component {
 
   onActivityTypeFilterChange(e) {
     const value = e.target.value.split(',')
+    const filteredActivity = e.target.value === 'all' ? [] : value
```
web-bert: I would have this as:
```javascript
const value = e.target.value
const filteredActivity = value === 'all' ? [] : value.split(',')
```

src/activity-feed/ActivityFeed.jsx
```diff
@@ -91,10 +90,18 @@ export default class ActivityFeed extends React.Component {
...
     } = this.state
 
+    const count = activities.filter((activity) => {
+      if (filteredActivity.length) {
+        return intersection(activity.object.type, filteredActivity).length
+      } else {
+        return true
+      }
+    }).length
+
```
paulgain: I realised this has been merged but Instead of using `lodash` we could have used the native `includes` on an Array. In addition we should of destructured the `activity` as well.  
```
const count = activities.filter(({ object }) => {
  return filteredActivity.length ? filteredActivity.includes(object.type).length : true
}).length
```

PR #213, opened 2019-10-24
refactor: deactivate activity type filter

yozzo:
deactivate activity type filter

### DIFF
--- a/src/activity-feed/ActivityFeed.jsx
+++ b/src/activity-feed/ActivityFeed.jsx
@@ -26,7 +26,7 @@ export default class ActivityFeed extends React.Component {
   static propTypes = {
     children: PropTypes.node,
     activities: PropTypes.arrayOf(PropTypes.object),
-    activityTypeFiltersList: PropTypes.array,
+    activityTypeFilters: PropTypes.array,
     onLoadMore: PropTypes.func,
     hasMore: PropTypes.bool,
     isLoading: PropTypes.bool,
@@ -38,7 +38,7 @@ export default class ActivityFeed extends React.Component {
   static defaultProps = {
     children: null,
     activities: [],
-    activityTypeFiltersList: [],
+    activityTypeFilters: [],
     onLoadMore: () => {},
     hasMore: false,
     isLoading: false,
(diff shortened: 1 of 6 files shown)


src/activity-feed/ActivityFeed.jsx
```diff
@@ -28,30 +26,36 @@ export default class ActivityFeed extends React.Component {
   static propTypes = {
     children: PropTypes.node,
     activities: PropTypes.arrayOf(PropTypes.object),
+    activityTypeFiltersList: PropTypes.array,
```
peterhudec: I think the `*List` suffix in the name is redundant. It is quite apparent that `activityTypeFilters` is an array. Also it would be more consistent with other props like `activities` etc.

src/activity-feed/ActivityFeed.jsx
```diff
@@ -28,30 +26,36 @@ export default class ActivityFeed extends React.Component {
...
     isLoading: PropTypes.bool,
     addContentText: PropTypes.string,
     addContentLink: PropTypes.string,
+    totalActivities: PropTypes.number,
   }
 
   static defaultProps = {
     children: null,
     activities: [],
+    activityTypeFiltersList: [],
```
peterhudec: Same as above, no need for the `*List` suffix.

src/activity-feed/constants.js
```diff
@@ -11,10 +11,8 @@ export const ACTIVITY_TYPE = {
   Omis: ['dit:OMISOrder'],
 }
 
-/**
- * To disable the Basic Activity Type filter, just return an empty array
- * ACTIVITY_TYPE_FILTERS = []
- */
+export const ACTIVE_ACTIVITY_TYPE_FILTERS = []
```
peterhudec: Do we really need this constant?

src/activity-feed/ActivityFeed.jsx
```diff
@@ -90,18 +95,10 @@ export default class ActivityFeed extends React.Component {
...
-      } else {
-        return true
-      }
-    }).length
-
     return (
       <ActivityFeedContainer>
         <ActivityFeedHeader
-          totalActivities={count}
+          totalActivities={totalActivities}
```
peterhudec: `totalActivities` is a default prop so you don't need to have this.

src/activity-feed/ActivityFeedApp.jsx
```diff
@@ -98,6 +99,7 @@ export default class ActivityFeedApp extends React.Component {
     return (
       <ActivityFeed
         addContentText={addContentText}
+        activityTypeFiltersList={ACTIVE_ACTIVITY_TYPE_FILTERS}
```
peterhudec: Also `activityTypeFiltersList` is a default prop so you don't need to have this.

PR #216, opened 2019-10-25
Feature/activity type filter

yozzo:
feat: activity type filter to query the api endpoint directly

Screenshot 2019-10-29 at 17 00 37


.yalc/data-hub-components/README.md
```diff
@@ -0,0 +1,142 @@
+# Data Hub Components
```
paulgain: We shouldn't be committing `.yalc` files. Instead we should add them to `.gitignore`.

.yalc/data-hub-components/package.json
```diff
@@ -0,0 +1,112 @@
+{
```
paulgain: We shouldn't be committing `.yalc` files. Instead we should add them to `.gitignore`.

.yalc/data-hub-components/yalc.sig
```diff
@@ -0,0 +1 @@
+c3e3abd5234a719307ca5b91fb509e9e
```
paulgain: We shouldn't be committing `.yalc` files. Instead we should add them to `.gitignore`.

.yalc/data-hub-components/README.md
yozzo: removed and added the files to `.git/info/exclude` as it's only a dev tool

paulgain: But doesn't that only exclude `./yalc` for you on your computer. What about everyone else?

paulgain: Surely it should be added to `.gitignore` in the same way our IDEs (dev tools) are added.

yozzo: added

package.json
```diff
@@ -25,8 +25,7 @@
     "storybook:build": "rimraf storybook-static && build-storybook --quiet",
     "storybook:release": "yarn storybook:build && cp -R ./.circleci ./storybook-static/.circleci && git-directory-deploy --directory storybook-static --branch gh-pages",
     "prepack": "yarn build",
-    "release": "semantic-release --debug",
-    "postpublish": "[ \"${npm_config_tag:-latest}\" != latest ] || yarn storybook:release"
+    "release": "semantic-release --debug"
   },
```
paulgain: Don't forget to put back the `postpublish` script otherwise we will not be able to release to `npm`.

src/activity-feed/ActivityFeed.jsx
```diff
@@ -26,49 +27,74 @@ export default class ActivityFeed extends React.Component {
...
+
+    /**
+     * the `key` param is used for building the query to ES
+     * "object.attributedTo.id" - for companies or advisers
+     * "object.type" - for activity types
+     */
+
+    if (
+      startsWith(filteredActivity[0], 'dit:DataHubAdviser:') ||
+      startsWith(filteredActivity[0], 'dit:DataHubCompany:')
```
paulgain: Shouldn't `dit:DataHubCompany` be `dit:Company`?

**Entire List**
0: "dit:Accounts"
1: "dit:Company"
2: "dit:Export"
3: "dit:Interaction"
4: "dit:ServiceDelivery"
5: "dit:InvestmentProject"
6: "dit:OMISOrder"

src/activity-feed/ActivityFeed.jsx
```diff
@@ -26,49 +27,74 @@ export default class ActivityFeed extends React.Component {
...
+          'object.attributedTo.id': filteredActivity,
+        },
+      ])
+    } else {
+      this.sendQueryParams([
+        {
+          'object.type': filteredActivity,
+        },
+      ])
+    }
```
paulgain: This can be simplified to:

    const queryParams = {}

    if (
      startsWith(filteredActivity[0], 'dit:DataHubAdviser:') ||
      startsWith(filteredActivity[0], 'dit:DataCompany:')
    ) {
      queryParams['object.attributedTo.id'] = filteredActivity
    } else {
      queryParams['object.type'] = filteredActivity
    }

    const { sendQueryParams } = this.props
    sendQueryParams([queryParams])

And this can be deleted:

     sendQueryParams(filteredActivity) {
        const { sendQueryParams } = this.props
        sendQueryParams(filteredActivity)
    }

src/activity-feed/ActivityFeed.jsx
yozzo: `dit:Company` refers to data coming from Companies House, in this case it represents external data

paulgain: I'm confused... the Select HTML looks like this:
```
  <label name="activity-types-filter" class="sc-fOICqy ghzlBB sc-jlyJG kBASKa">
    <span class="sc-gipzik hsjUYR">Activity types</span>
    <select class="sc-Rmtcm eBAkhk">
      <option value="dit:Accounts,dit:Company,dit:Export,dit:Interaction,dit:ServiceDelivery,dit:InvestmentProject,dit:OMISOrder">
        All Data Hub &amp; external activity
      </option>
      <option value="dit:DataHubAdviser:123-456">
        My activity
      </option>
      <option value="dit:Accounts,dit:Company,dit:Export">
        All external activity
      </option>
      <option value="dit:Interaction,dit:ServiceDelivery,dit:InvestmentProject,dit:OMISOrder" selected="">
        All Data Hub activity
      </option>
    </select>
  </label>
```
`dit:DataHubCompany ` doesn't exist within the HTML, therefore, `|| startsWith(filteredActivity[0], 'dit:DataHubCompany:')` will always be `false`. What am I missing?

yozzo: as agreed verbally these keys are used for choosing between `object.type` and `object.attributedTo` query param keys.

src/activity-feed/constants.js
```diff
@@ -36,7 +61,7 @@ export const ACTIVITY_TYPE_FILTERS = [
       ]
     ),
   },
-]
+}
```
paulgain: Theses constants can be improved to:  

    export const ACTIVITY_TYPE_FILTERS = {
      allActivity: {
        label: 'All Data Hub & external activity',
        value: [
          ...ACTIVITY_TYPE.CompaniesHouseAccount,
          ...ACTIVITY_TYPE.CompaniesHouseCompany,
          ...ACTIVITY_TYPE.HmrcExporter,
          ...ACTIVITY_TYPE.Interaction,
          ...ACTIVITY_TYPE.InvestmentProject,
          ...ACTIVITY_TYPE.Omis,
        ],
      },
      myActivity: {
        label: 'My activity',
        value: 'dit:DataHubAdviser:123-456',
      },
      externalActivity: {
        label: 'All external activity',
        value: [
          ...ACTIVITY_TYPE.CompaniesHouseAccount,
          ...ACTIVITY_TYPE.CompaniesHouseCompany,
          ...ACTIVITY_TYPE.HmrcExporter,
        ],
      },
      dataHubActivity: {
        label: 'All Data Hub activity',
        value: [
          ...ACTIVITY_TYPE.Interaction,
          ...ACTIVITY_TYPE.InvestmentProject,
          ...ACTIVITY_TYPE.Omis,
        ],
      },
    }

We can then access them like:
  
    const = { allActivity, myActivity, externalActivity, dataHubActivity } = ACTIVITY_TYPE_FILTERS

yozzo: that's how I declared them initially, but then we'd still need to add them in an array and pass them to the `<Select/>` as such, and also define the `default.value`. Whereas right now if we need to add, remove, or change a default a filter value then it can be done easily only here without needing to touch the code anywhere else.

paulgain: 1. They're already in an array due to the `spread operator`. 
2. The default value can be whatever you choose, for example:

    queryParams: [{ 'object.type': ACTIVITY_TYPE_FILTERS. dataHubActivity.value }]

There's no need to pass all of these `array` values to the `select`, the `select` should simply have 4 `strings`, where each `string` is the value of the `option`, each string represents one of the objects above. After a user has made a selection you map them within the event handler and then build the query. There's really no need to shove all this query data into a `select`.

src/activity-feed/ActivityFeed.jsx
```diff
@@ -26,49 +27,80 @@ export default class ActivityFeed extends React.Component {
...
   }
 
   static defaultProps = {
     children: null,
     activities: [],
-    activityTypeFilters: [],
+    activityTypeFilters: {},
     onLoadMore: () => {},
     hasMore: false,
+    isFilterEnabled: false,
```
rafenden: This prop should be called `isTypeFilterEnabled` as it enables only the activity type filter.

paulgain:
@yozzo that's a lot cleaner now 👍

PR #217, opened 2019-10-28
feat: Add Badge component

rafenden:
It will be used on the activity feed and on the collection pages.

image

### DIFF
--- a/src/badge/Badge.jsx
+++ b/src/badge/Badge.jsx
@@ -1,0 +1,34 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { FONT_SIZE } from '@govuk-react/constants'
+import VisuallyHidden from '@govuk-react/visually-hidden'
+import { GREY_2 } from 'govuk-colours'
+
+const StyledBadge = styled('span')`
+  border: ${({ borderColour }) => `2px solid ${borderColour}`};
+  border-radius: 4px;
+  padding: 2px 4px;
+  font-size: ${FONT_SIZE.SIZE_14};
+  white-space: nowrap;
+`
+
+const Badge = ({ label, borderColour, children }) => (
+  <StyledBadge borderColour={borderColour}>
+    {label && <VisuallyHidden>{label}</VisuallyHidden>}
+    {children}
+  </StyledBadge>
+)
+
+Badge.propTypes = {
+  label: PropTypes.string,
+  borderColour: PropTypes.string,
+  children: PropTypes.node.isRequired,
+}
+
+Badge.defaultProps = {
+  label: null,
+  borderColour: GREY_2,
+}
+
+export default Badge
--- a/src/badge/__tests__/Badge.test.jsx
+++ b/src/badge/__tests__/Badge.test.jsx
@@ -1,0 +1,55 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import VisuallyHidden from '@govuk-react/visually-hidden'
+import Badge from '../Badge'
+
+describe('Badge', () => {
+  let wrapper
+
+  describe('when no props were passed', () => {
+    beforeAll(() => {
+      wrapper = mount(<Badge>testChildren</Badge>)
+    })
+
+    test('should set the content with children', () => {
+      expect(wrapper.text()).toEqual('testChildren')
+    })
+
+    test('should not render a hidden label', () => {
+      expect(wrapper.find(VisuallyHidden)).toHaveLength(0)
+    })
+
+    test('should render with a default (gray) border', () => {
+      expect(wrapper).toHaveStyleRule('border', '2px solid #bfc1c3')
+    })
+  })
+
+  describe('when a "label" prop is passed', () => {
+    beforeAll(() => {
+      wrapper = mount(<Badge label="testLabel">testChildren</Badge>)
+    })
+
+    test('should render a hidden label', () => {
+      expect(wrapper.find(VisuallyHidden).text()).toEqual('testLabel')
+    })
+
+    test('should set the content with children', () => {
+      expect(wrapper.text()).toContain('testChildren')
+    })
+  })
+
+  describe('when a "borderColour" prop is passed', () => {
+    beforeAll(() => {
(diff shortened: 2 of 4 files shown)


src/badge/Badge.jsx
```diff
@@ -0,0 +1,38 @@
...
+const StyledBadge = styled('span')`
+  border: ${({ borderColour }) => `2px solid ${borderColour}`};
+  border-radius: 4px;
+  padding: 2px 4px;
+  font-size: ${FONT_SIZE.SIZE_14};
+  white-space: nowrap;
+  margin-right: ${SPACING.SCALE_1};
+  &:last-child {
+    margin-right: 0;
+  }
```
peterhudec: It's better to not control outer layout from inside of a component. This is the responsibility of the parent. Look for example how Material UI's [chips](https://material-ui.com/components/chips/) don't have any margin exactly for this reason.

src/badge/__tests__/Badge.test.jsx
```diff
@@ -0,0 +1,55 @@
...
+import VisuallyHidden from '@govuk-react/visually-hidden'
+import Badge from '../Badge'
+
+describe('Badge', () => {
+  let wrapper
+
+  describe('when no props were passed', () => {
+    beforeAll(() => {
+      wrapper = mount(<Badge>testChildren</Badge>)
+    })
```
peterhudec: The setup and teardown hooks are to isolate side-effecting tests. You don't need to re-mount the `Badge` between each assertion. They are all part of the same test case.

src/badge/__tests__/Badge.test.jsx
```diff
@@ -0,0 +1,55 @@
...
+      wrapper = mount(<Badge label="testLabel">testChildren</Badge>)
+    })
+
+    test('should render a hidden label', () => {
+      expect(wrapper.find(VisuallyHidden).text()).toEqual('testLabel')
+    })
+
+    test('should set the content with children', () => {
+      expect(wrapper.text()).toContain('testChildren')
+    })
```
peterhudec: Same here. This is in fact a single test with two assertions.
```jsx
test('label prop', () => {
  const wrapper = mount(<Badge label="testLabel">testChildren</Badge>)
  expect(wrapper.find(VisuallyHidden).text()).toEqual('testLabel')
  expect(wrapper.text()).toContain('testChildren')
})
```

src/badge/Badge.jsx
rafenden: I check the code for Material UI chips and it looks like they do specify margin:
https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/chips/Chips.js#L14

Do you refer to the right margin only?

peterhudec: The link is the actual example from the docs about how to use `Chip`. The actual component implemented in the example is called `Chips` and it is setting the spacing for the `Chips` it is using. This is actually exactly how it should be.

src/badge/__tests__/Badge.test.jsx
rafenden: > You don't need to re-mount the Badge between each assertion.

`beforeAll` renders the component only once.

Tests in component library and the frontend repo were written using BDD style with preparation separated from the assertion.

Happy to discuss how we could change the way we write tests in general, but it's outside the scope for this PR.

src/badge/Badge.jsx
rafenden: Now I get it, the margin was removed now.

src/badge/__tests__/Badge.test.jsx
rafenden: In this case, separated assertions are more explicit making it clear what's exactly

src/badge/__tests__/Badge.test.jsx
peterhudec: Yeah, this requires broader discussion. I've added a section about it to the [Testing Strategy for Data Hub Frontend](https://docs.google.com/document/d/1ltHFzDaA57RCrRcymFGSJE_L9x8W7LAxnRE4InAMNXM/edit?ts=5da83cc6#heading=h.6969vxk9vn4p) document.

src/badge/Badge.jsx
```diff
@@ -0,0 +1,34 @@
...
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { FONT_SIZE } from '@govuk-react/constants'
+import VisuallyHidden from '@govuk-react/visually-hidden'
+import { GREY_2 } from 'govuk-colours'
+
+const StyledBadge = styled('span')`
+  border: ${({ borderColour }) => `2px solid ${borderColour}`};
+  border-radius: 4px;
+  padding: 2px 4px;
```
debitan: I didn't realise when I wrote my badge component before, but we can swap most of these `px` sizings for constants. 

`4px` is equal to `BORDER_WIDTH_MOBILE` and `2px` is equal to `BORDER_WIDTH_FORM_ELEMENT`

src/badge/Badge.jsx
```diff
@@ -0,0 +1,34 @@
...
+  font-size: ${FONT_SIZE.SIZE_14};
+  white-space: nowrap;
+`
+
+const Badge = ({ label, borderColour, children }) => (
+  <StyledBadge borderColour={borderColour}>
+    {label && <VisuallyHidden>{label}</VisuallyHidden>}
+    {children}
```
debitan: Will the label ever be different from the children? If not, could we pass the text contained within children to the `<VisuallyHidden />` tag to save passing in the addition `label` argument? 

e.g.
```
{children && <VisuallyHidden>{children}</VisuallyHidden>}
{children}
```

src/badge/Badge.jsx
rafenden: It doesn't feel right to use constants with misleading names. We should probably have a separate constants file where we reuse them across components.

src/badge/Badge.jsx
rafenden: In most cases label should be different, e.g.

```
<Badge label="Headquarter type">Global HQ</Badge>
```

Without the label user will not know what `Global HQ` refers to.

rafenden:
> do we need to export this as well from index.jsx?
Good point, will change the code.

PR #218, opened 2019-10-28
feat: Add Metadata component

rafenden:
Add component for metadata which will be used on collection pages, activity feed and in a few more places.

image

### DIFF
--- a/src/metadata/Metadata.jsx
+++ b/src/metadata/Metadata.jsx
@@ -1,0 +1,42 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { FONT_SIZE, SPACING } from '@govuk-react/constants'
+
+import MetadataItem from './MetadataItem'
+
+const StyledMetadataWrapper = styled('div')`
+  font-size: ${FONT_SIZE.SIZE_16};
+  line-height: ${FONT_SIZE.SIZE_27};
+  display: grid;
+
+  & > * {
+    margin-bottom: ${SPACING.SCALE_1};
+  }
+`
+
+const Metadata = ({ rows }) =>
+  rows && (
+    <StyledMetadataWrapper>
+      {rows.map(({ label, value }) => (
+        <MetadataItem key={label} label={label}>
+          {value}
+        </MetadataItem>
+      ))}
+    </StyledMetadataWrapper>
+  )
+
+Metadata.propTypes = {
+  rows: PropTypes.arrayOf(
+    PropTypes.shape({
+      label: PropTypes.node,
+      value: PropTypes.node,
+    })
+  ),
+}
+
+Metadata.defaultProps = {
+  children: null,
+}
+
+export default Metadata
--- a/src/metadata/MetadataItem.jsx
+++ b/src/metadata/MetadataItem.jsx
@@ -1,0 +1,31 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { BLACK, GREY_1 } from 'govuk-colours'
+
+const StyledMetaWrapper = styled('div')`
+  color: ${BLACK};
+`
+
+const StyledItemLabel = styled('span')`
+  color: ${GREY_1};
+`
+
+function MetadataItem({ label, children }) {
+  return (
+    <StyledMetaWrapper>
+      {label && <StyledItemLabel>{label}</StyledItemLabel>} {children}
+    </StyledMetaWrapper>
+  )
+}
+
+MetadataItem.propTypes = {
+  label: PropTypes.string,
+  children: PropTypes.node.isRequired,
+}
+
+MetadataItem.defaultProps = {
+  label: null,
+}
+
+export default MetadataItem
--- a/src/metadata/__tests__/Metadata.test.jsx
+++ b/src/metadata/__tests__/Metadata.test.jsx
@@ -1,0 +1,41 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import Link from '@govuk-react/link'
+import Metadata from '../Metadata'
+
+describe('Metadata', () => {
+  let wrapper
+
+  describe('when "rows" prop was passed with no values', () => {
+    beforeAll(() => {
+      wrapper = mount(<Metadata rows={null} />)
+    })
+
+    test('should not render anything', () => {
+      expect(wrapper.html()).toEqual(null)
+    })
+  })
+
+  describe('when a "rows" prop is passed with values', () => {
+    beforeAll(() => {
+      wrapper = mount(
+        <Metadata
+          rows={[
+            { label: 'Updated on', value: '5 September 2019' },
+            {
+              label: 'Parent company',
+              value: <Link href="http://example.com">E-corp LTD</Link>,
+            },
+          ]}
+        />
+      )
+    })
+
+    test('should render all metadata items', () => {
+      expect(wrapper.text()).toContain(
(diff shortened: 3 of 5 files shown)


src/metadata/MetadataItem.jsx
```diff
@@ -0,0 +1,33 @@
...
+import { SPACING } from '@govuk-react/constants'
+import { BLACK, GREY_1 } from 'govuk-colours'
+
+const StyledMetaWrapper = styled('span')`
+  margin-top: ${SPACING.SCALE_1};
+  color: ${BLACK};
+`
+
+const StyledItemLabel = styled('span')`
+  color: ${GREY_1};
```
peterhudec: Why are these spans?

src/metadata/Metadata.jsx
```diff
@@ -0,0 +1,43 @@
...
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { FONT_SIZE, MEDIA_QUERIES, SPACING } from '@govuk-react/constants'
+
+import MetadataItem from './MetadataItem'
+
+const StyledMetadataWrapper = styled('div')`
+  font-size: ${FONT_SIZE.SIZE_16};
+  line-height: ${FONT_SIZE.SIZE_27};
+  margin-top: ${SPACING.SCALE_1};
```
peterhudec: Why should this component care about space around itself? This is the concern of its parent.

src/metadata/Metadata.jsx
```diff
@@ -0,0 +1,43 @@
...
+  display: grid;
+
+  ${MEDIA_QUERIES.TABLET} {
+    margin-top: -${SPACING.SCALE_3};
+  }
```
peterhudec: Same as above, this should be handled by whoever is using `Metadata`.

src/metadata/__tests__/Metadata.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+import Link from '@govuk-react/link'
+import Metadata from '../Metadata'
+
+describe('Metadata', () => {
+  let wrapper
+
+  describe('when "rows" prop was passed with no values', () => {
+    beforeAll(() => {
+      wrapper = mount(<Metadata rows={null} />)
+    })
```
peterhudec: Do you really need `beforeAll` when you only have one test in the `describe` block? Also you can just ommit `rows`.
```jsx
test('should not render anything', () =>
  expect(wrapper.html(mount(<Metadata />)).toEqual(null)
)
```

src/metadata/__tests__/Metadata.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+    test('should not render anything', () => {
+      expect(wrapper.html()).toEqual(null)
+    })
+  })
+
+  describe('when a "rows" prop is passed with values', () => {
+    beforeAll(() => {
```
peterhudec: Same here. You don't need `beforeAll`.

src/metadata/__tests__/Metadata.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+              value: <Link href="http://example.com">E-corp LTD</Link>,
+            },
+          ]}
+        />
+      )
+    })
+
+    test('should render all metadata items', () => {
+      expect(wrapper.text()).toContain('Updated on 5 September 2019')
+      expect(wrapper.text()).toContain('Parent company E-corp LTD')
```
peterhudec: This would not fail if the rows were in different order, but maybe tha'ts the desired behavior. If not, then better assert the whole text.

peterhudec: I find it much more readable without the wrapper variable:
```jsx
describe('expect(wrapper.html(mount(<Metadata />)).toEqual(null)', () =>
  test('should render all metadata items', () =>
    expect(mount(
      <Metadata
        rows={[
          { label: 'Updated on', value: '5 September 2019' },
          {
            label: 'Parent company',
            value: <Link href="http://example.com">E-corp LTD</Link>,
          },
        ]}
      />
    ).toContain(
      'Updated on 5 September 2019' +
      'Parent company E-corp LTD'
    )
  )
)
```

src/metadata/__tests__/Metadata.test.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import Link from '@govuk-react/link'
+import Metadata from '../Metadata'
+
+describe('Metadata', () => {
+  let wrapper
```
peterhudec: All the tests share this variable, even though, there's no and there should be no connection between the tests. If you forgot to reassign it, you could easily end up making assertion about a wrapper from a previous test. You actually don't need to assign the wrapper at all.

src/metadata/MetadataItem.jsx
rafenden: To change colour of the label.

src/metadata/__tests__/Metadata.test.jsx
rafenden: It's a common pattern among all tests.

src/metadata/Metadata.jsx
rafenden: Good point, margin is now removed.

src/metadata/Metadata.jsx
rafenden: Good point, margin is now removed.

src/metadata/__tests__/Metadata.test.jsx
rafenden: It's a pattern, done for consistency.

src/metadata/__tests__/Metadata.test.jsx
rafenden: IMO, separation of assertions makes the test readable and easier to reason about, happy to chat about this offline.

src/metadata/MetadataItem.jsx
peterhudec: Sorry, the question was supposed to be on the `StyledMetaWrapper` component, which was a span, but you have changed it to a div since.

src/metadata/Metadata.jsx
```diff
@@ -0,0 +1,42 @@
...
+import MetadataItem from './MetadataItem'
+
+const StyledMetadataWrapper = styled('div')`
+  font-size: ${FONT_SIZE.SIZE_16};
+  line-height: ${FONT_SIZE.SIZE_27};
+  display: grid;
+
+  & > * {
+    margin-bottom: ${SPACING.SCALE_1};
+  }
```
debitan: I've been trying to avoid the SASS style selectors and explicitly write the style for the child components. For me, at least, it is easier to read and understand what is going on. 

I think this `margin-bottom` belongs on the MetadataItem component.

debitan: This has lost the 

```
${MEDIA_QUERIES.TABLET} {
    margin-top: -${SPACING.SCALE_3};
  }
```

styling I had in my metadata component. This means there is a big gap between the item title and the metadata when I try to use this component in my CollectionItem. I also had `margin: ${SPACING.SCALE_2} 0;`, but that doesn't seem to be an obvious issue.

rafenden: It was in `MetadataItem` before but after request from @peterhudec it was moved to the parent, I think it's reasonable to have it there.

rafenden: This style will be added to the `CollectionItem` component as it's not related to the `Metadata`.

PR #219, opened 2019-10-31
feat: Add Pagination component

rafenden:
This component was extracted from the collection list and will replace it in a follow-up PR.

image

### DIFF
--- a/src/pagination/Pagination.jsx
+++ b/src/pagination/Pagination.jsx
@@ -1,0 +1,159 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { GREY_1, GREY_3, GREY_4, LINK_COLOUR, TEXT_COLOUR } from 'govuk-colours'
+import { FONT_SIZE, MEDIA_QUERIES, SPACING } from '@govuk-react/constants'
+import Link from '@govuk-react/link'
+
+import computeVisiblePieces from './computeVisiblePieces'
+
+import {
+  PAGINATION_PIECE_ELLIPSIS,
+  PAGINATION_PIECE_NEXT,
+  PAGINATION_PIECE_PAGE_NUMBER,
+  PAGINATION_PIECE_PREVIOUS,
+} from './constants'
+
+const StyledNav = styled('nav')`
+  text-align: center;
+  line-height: 1;
+  display: flex;
+  justify-content: space-around;
+  padding: ${SPACING.SCALE_3};
+
+  ${MEDIA_QUERIES.TABLET} {
+    display: block;
+  }
+`
+
+const StyledPaginationList = styled('ul')`
+  margin: 0;
+  list-style: none;
+`
+
+const StyledPaginationPiece = styled('li')`
+  display: inline-block;
+
+  & + & {
+    margin-left: ${SPACING.SCALE_1};
+  }
+`
+
+const StyledPaginationLink = styled(Link)`
+  font-weight: bold;
+  font-size: ${FONT_SIZE.SIZE_16};
+  display: inline-block;
+  padding: ${SPACING.SCALE_1} ${SPACING.SCALE_3};
+  background-color: ${GREY_4};
+  line-height: 1.9em;
+  color: ${LINK_COLOUR};
+  text-decoration: none;
+
+  :hover {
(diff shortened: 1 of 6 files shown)


src/pagination/__tests__/Pagination.test.jsx
```diff
@@ -0,0 +1,203 @@
+import React from 'react'
+import { mount } from 'enzyme/build'
+import Pagination from '../Pagination'
+
+describe('Pagination', () => {
+  let wrapper
+
+  describe('when 7 pages are passed and the active page is 3', () => {
+    beforeAll(() => {
```
peterhudec: No need for `beforeAll` as all the tests only differ by assertions. It's actually a single test with multiple assertions. The "should render the ..." messages should actually be custom assertion messages. Unfortunately, [Jest doesn't support this standard feature out of the box](https://github.com/facebook/jest/issues/3293), but the [`jest-expect-message`](https://www.npmjs.com/package/jest-expect-message) package solves this problem.

Apart from the tests to appear more complex than they really are, mounting the same stateless component 5 times just to make 5 assertions can affect the duration of the whole test suite. Especially if we do it everywhere, which we do. Just in this PR there are 24 `mount` calls, when it only could be 7.

src/pagination/__tests__/Pagination.test.jsx
```diff
@@ -0,0 +1,203 @@
...
+
+    test('should render the truncated page link', () => {
+      expect(wrapper.find('li span')).toHaveLength(1)
+    })
+
+    test('should render the active page', () => {
+      expect(wrapper.find('a[data-test="page-number-active"]').text()).toEqual(
+        '3'
+      )
+    })
```
peterhudec: All these are assertions and not tests. The `describe` block is the actual test. It should be:
```jsx
test('when 7 pages are passed and the active page is 3', () => {
  const wrapper = mount(<Pagination activePage={3} totalPages={7} />)

  expect(
    wrapper
      .find('a')
      .first()
      .text(),
    // This would work only with jest-expect-message
    'should render the previous button',
  ).toBe('Previous')
  
  expect(
    wrapper
      .find('a')
      .last()
      .text(),
    'should render the next button',
  ).toBe('Next')

  expect(
    wrapper.find('li'),
    'should render all the pagination pieces',
  ).toHaveLength(9)

  expect(
    wrapper.find('li span'),
    'should render the truncated page link',
  ).toHaveLength(1)

  expect(
    wrapper.find('a[data-test="page-number-active"]').text(),
    'should render the active page',
  ).toEqual('3')
})
```

src/pagination/Pagination.jsx
```diff
@@ -0,0 +1,166 @@
...
+                  onClick={onClick}
+                  href={getPageUrl(pageNumber)}
+                >
+                  Previous
+                </StyledPrevious>
+              )}
+
+              {type === PAGINATION_PIECE_ELLIPSIS && (
+                <StyledEllipsis key={key} data-test="ellipsis">
+                  …
```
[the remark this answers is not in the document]
rafenden: I would prefer to avoid extra and unnecessary component.

src/pagination/__tests__/Pagination.test.jsx
rafenden: The component it's mounted only once as `beforeAll` is executed once for all the test cases.

src/pagination/__tests__/Pagination.test.jsx
rafenden: As I mentioned in a previous PR, those tests are written in BDD style, as the rest of the codebase.

src/pagination/Pagination.jsx
```diff
@@ -0,0 +1,159 @@
...
+  margin: 0;
+  list-style: none;
+`
+
+const StyledPaginationPiece = styled('li')`
+  display: inline-block;
+
+  & + & {
+    margin-left: ${SPACING.SCALE_1};
+  }
```
debitan: We should add this to the adjacent component rather than using SASS selectors.

rafenden: It was changed following comment added by @peterhudec https://github.com/uktrade/data-hub-components/pull/218#discussion_r339634713

PR #226, opened 2019-11-21
feat: Add Ultimate HQ subsidiaries filter

paulgain:
## Description of change
When viewing a D&B Ultimate HQ company a user is now able to filter all subsidiary activities across the entire D&B Ultimate HQ hierarchy when clicking the filter box `Activities across all x companies`. For example, If you have three companies in a D&B hierarchy:

1. Ultimate HQ (has 3 activities)
2. Subsidiary A (has 2 activities)
3. Subsidiary B  (has 2 activities)

When applying the filter the Activity Feed list will display all 7 activities across this hierarchy. The filter works in conjunction with the following Activity types.

**Activity types**
1. `All Data Hub & external activity`
2. `All external activity`
3. `My activity`
4. `All Data Hub activity`

This functionality is hidden behind a feature flag `companies-ultimate-hq`

ultimate-hq-subsidiaries-filter


src/activity-feed/ActivityFeed.jsx
```diff
@@ -60,8 +67,10 @@ export default class ActivityFeed extends React.Component {
     } = props.activityTypeFilters
 
     super(props)
+
     this.state = {
-      filteredActivity: dataHubActivity ? dataHubActivity.value : '',
+      activityTypeFilter: dataHubActivity ? dataHubActivity.value : '',
+      showDnBHierarchy: false,
```
web-bert: This should be `showDnbHierarchy` to be consistent camelCase and as used for `dnbHierachyCount`

paulgain:
> Just one comment about camelCase for now, but this is above our agreed line limit for a PR and I feel like I need to take some time to review properly, which I don't feel I have at the moment. @mcbhenwood thoughts?

Apologies for the refactor, however, I've significantly improved what was there, unfortunately the code was becoming unmanageable. I can take you through the code which will save some time. There is also a corresponding frontend piece that has already been approved.

src/activity-feed/ActivityFeed.jsx
paulgain: 👍

PR #231, opened 2019-12-13
feat: Add a static function - formattedDateAndTime

paulgain:
Added a simple helper function that will take a date and time such as: `2019-12-10T18:12:37.649476Z` and format it to: `10 Dec 2019, 6:12pm`

### DIFF
--- a/src/utils/DateUtils.js
+++ b/src/utils/DateUtils.js
@@ -4,4 +4,8 @@ export default class DateUtils {
   static format(dateStr) {
     return moment(dateStr).format('DD MMM YYYY')
   }
+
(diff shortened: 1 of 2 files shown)


src/utils/DateUtils.js
```diff
@@ -4,4 +4,8 @@ export default class DateUtils {
   static format(dateStr) {
     return moment(dateStr).format('DD MMM YYYY')
   }
+
+  static updatedOn(dateStr) {
```
rafenden: Probably, we should call it `formatWithTime` and use it for all dates with time.

paulgain: Makes sense, thanks.

PR #236, opened 2020-01-06
ci: Remove dependabot config so that reviewers can be added in dependabot dashboard

rafenden:


.dependabot/config.yml
```diff
@@ -11,4 +11,9 @@ update_configs:
     default_reviewers:
       - paulgain
       - rafenden
-      - yeahfro
+      - debitan
+      - ian-leggett
+      - paulgain
+      - peterhudec
+      - web-bert
+      - yozzo
```
web-bert: @rafenden can this be the @uktrade/datahub-front-end  team?

paulgain: @rafenden I'm on the list twice.

rafenden: > @rafenden can this be the @uktrade/datahub-front-end team?
Good idea, will update the code.

rafenden: Unfortunately, this is currently not supported 😢 
https://github.com/dependabot/feedback/issues/546

rafenden: Unfortunately, this is currently not supported 😢
dependabot/feedback#546

PR #237, opened 2020-01-07
feat: Add USD currency formatter

paulgain:
Added a USD currency formatter.


paulgain:
> This is a breaking change, please add a relevant message to the commit body.
> 
> e.g.
> 
> ```
> fix: Remove ActivityFeedApp component
> 
> BREAKING CHANGE: This components is no longer used so we removed it.
> Optionally add more info in the second line of your message.
> ```

👍

PR #240, opened 2020-01-13
feat: Show an H3 header if no link is passed to the collectionItem

debitan:
# Before

before

# After

after

### DIFF
--- a/src/collection-list/__tests__/CollectionItem.test.jsx
+++ b/src/collection-list/__tests__/CollectionItem.test.jsx
@@ -100,7 +100,7 @@ describe('CollectionItem', () => {
     })
 
     test('should not render the headingUrl', () => {
-      expect(wrapper.find('a').exists()).toBe(false)
+      expect(wrapper.find('a')).toHaveLength(0)
     })
   })
 })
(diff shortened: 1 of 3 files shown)


src/collection-list/__tests__/CollectionItem.test.jsx
```diff
@@ -79,4 +79,28 @@ describe('CollectionItem', () => {
...
+    test('should render the component', () => {
+      expect(wrapper.find(CollectionItem).exists()).toBe(true)
+    })
+
+    test('should render the headingText', () => {
+      expect(wrapper.find('h3').text()).toBe('Mars Exports Ltd')
+    })
+
+    test('should not render the headingUrl', () => {
+      expect(wrapper.find('a').exists()).toBe(false)
```
rafenden: One more option to write this assetion:
```suggestion
      expect(wrapper.find('a')).toHaveLength(0)
```

debitan: do we have a preference? I've used `exists()).toBe` higher up in the test file, so it is consistent. is one more performant/better than the other?

rafenden: I personally tend to use `expect(wrapper.find('a')).toHaveLength(0)` as it uses fewer characters and is easier to read.

Can't comment on the performance.

PR #241, opened 2020-01-14
feat: hide collectionItem metadata behind details if > 4

debitan:
Hide `collectionItem` metadata behind details if > 4. Add optional prop `type` to pass to the details summary message. If no `type` prop is passed the message will display 'View details'.

# Less than 5 items

before

# More than 5 items

after2

# More than 5 items (expanded)

after1

### DIFF
--- a/src/collection-list/CollectionItem.jsx
+++ b/src/collection-list/CollectionItem.jsx
@@ -6,6 +6,7 @@ import { H3 } from '@govuk-react/heading'
 import Link from '@govuk-react/link'
 import { HEADING_SIZES, SPACING } from '@govuk-react/constants'
 import { GREY_2 } from 'govuk-colours'
+import Details from '@govuk-react/details'
 
 import Metadata from '../metadata/Metadata'
 import Badge from '../badge/Badge'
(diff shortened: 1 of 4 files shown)


src/collection-list/CollectionItem.jsx
```diff
@@ -67,7 +75,15 @@ function CollectionItem({
 
       <StyledSubheading>{subheading}</StyledSubheading>
 
-      <Metadata rows={metadata} />
+      {metadata && metadata.length > 4 ? (
```
web-bert: Can this be when the delta is 2 or more as there is no point putting this behind a disclosure if the total is 5. I would make a `const MIN_ITEMS = 4` and then do `if((metadata.length - MIN_ITEMS) > 2)`

debitan: I'm not sure that complexity is necessary here as it is either show all or hide everything and 5 is the exact minimum number of items an interaction can have, and we want to hide it.

web-bert: If we want this to be a generic component then it should have some logic like this, as a user of this you'll end up getting no disclosure when the list is 4 long, but when it gets to 5 it will be - then it might as well have just shown the 5th item. i.e. it's not saving any visual space. To make this work for interactions, the `MIN_ITEMS` should be 3

PR #242, opened 2020-01-14
feat: allow CollectionItem to pass border-colour to the badge component

debitan:
Refactor the CollectionItem component to allow it to pass a `border-colour` to the badge component.

## No colour passed (defaults to `GREY_2`)

noColour

## Green passed 

colour

### DIFF
--- a/src/collection-list/CollectionItem.jsx
+++ b/src/collection-list/CollectionItem.jsx
@@ -66,7 +66,9 @@ function CollectionItem({
       {badges && (
         <StyledBadgesWrapper>
           {badges.map((badge) => (
-            <Badge key={badge}>{badge}</Badge>
+            <Badge key={badge.text} borderColour={badge.borderColour}>
+              {badge.text}
+            </Badge>
           ))}
         </StyledBadgesWrapper>
       )}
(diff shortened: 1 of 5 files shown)


src/collection-list/CollectionItem.jsx
```diff
@@ -82,7 +84,7 @@ CollectionItem.propTypes = {
   headingUrl: PropTypes.string,
   headingText: PropTypes.string.isRequired,
   subheading: PropTypes.string,
-  badges: PropTypes.arrayOf(PropTypes.string),
+  badges: PropTypes.arrayOf(PropTypes.object),
```
paulgain: It would be helpful to define what these objects are comprised of.

debitan: with shape, I presume?

paulgain: Absolutely 👍

debitan: @paulgain it's done now. Green ticket please 🙇

paulgain: :boom:

rafenden:
@debitan aren't we going away from colours in badges as they are not easy to distinguish?

debitan:
> @debitan aren't we going away from colours in badges as they are not easy to distinguish?

The design for the export markets history page pulls in interactions, so they are to be coloured green the same as the activity feed. I've double checked this with Farhad and Rogier. I've got another PR waiting after this one to refactor the transformer for your hierarchy page that will shape the data in the new way so as not to break the site (!). The other implementation, large capital profiles, doesn't show badges at all - so no work needed there.

PR #246, opened 2020-01-21
feat: Make the premature form exit prompt configurable

rafenden:
Sometimes we might not need the prompt when user exists a form without submitting.


rafenden:
> Is it worth getting a test around this?

That's the tricky part, couldn't find out how to test the `beforeunload` event.

ian-leggett:
> > Is it worth getting a test around this?
> 
> That's the tricky part, couldn't find out how to test the `beforeunload` event with enzyme.

No worries we can catch this stuff in functionals.

rafenden:
> No worries we can catch this stuff in functionals.

@ian-leggett Looks like I found a way, have looks at the updated code 😄

PR #250, opened 2020-02-03
Typeahead

ian-leggett:
## Description
A new typeahead that we need in the referrals form, this will also eventually replace all the Vue typeaheads.

Trello - https://trello.com/c/xgH5rrE0/1095-build-react-typeahead-component

Note: We are only decorating a new dependancy (React Select) - https://react-select.com/home so I only have tests around the highlighting of matched characters as you type. We will test this in depth once ported to FE.

### DIFF
--- a/src/typeahead/Highlighter.jsx
+++ b/src/typeahead/Highlighter.jsx
@@ -1,0 +1,24 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import reactStringReplace from 'react-string-replace'
+import styled from 'styled-components'
+import { FONT_WEIGHTS } from '@govuk-react/constants'
+
+const StyledSpan = styled('span')`
+  font-weight: ${FONT_WEIGHTS.bold};
+`
+
+const Highlighter = ({ optionLabel = null, searchStr }) =>
+  optionLabel &&
+  reactStringReplace(optionLabel, searchStr, (matchedTxt, i) => (
+    <StyledSpan key={i} data-test-id={`test-${i}`}>
+      {matchedTxt}
+    </StyledSpan>
+  ))
+
+Highlighter.propTypes = {
+  optionLabel: PropTypes.string.isRequired,
+  searchStr: PropTypes.string.isRequired,
+}
+
+export default Highlighter
--- a/src/typeahead/Typeahead.jsx
+++ b/src/typeahead/Typeahead.jsx
@@ -1,0 +1,82 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Select, { components } from 'react-select'
+import AsyncSelect from 'react-select/async'
+import defaultStyles from './styles'
+import Highlighter from './Highlighter'
+
+const Option = ({ inputValue, ...props }) => {
+  const {
+    data: { label: optionLabel },
+  } = props
+
+  return (
+    <components.Option {...props}>
+      {inputValue ? (
+        <Highlighter searchStr={inputValue} optionLabel={optionLabel} />
+      ) : (
+        optionLabel
+      )}
+    </components.Option>
+  )
+}
+
+const Typeahead = ({ name, value, options, inputValue, styles, ...rest }) => {
+  const commonProps = {
+    styles: {
+      ...defaultStyles,
+      ...styles,
+    },
+    components: {
+      Option: (props) => <Option {...props} inputValue={inputValue} />,
+    },
+  }
+
+  const filterOptions = ({ label: labelToFilter }) =>
+    labelToFilter ? labelToFilter.toLowerCase().includes(inputValue) : false
+
+  return (
+    <>
+      {options ? (
+        <Select
+          {...commonProps}
+          defaultValue={options.filter((option) => option.value === value)}
+          options={options}
+          {...rest}
+        />
+      ) : (
+        <AsyncSelect
+          {...commonProps}
+          filterOption={filterOptions}
+          defaultValue={value}
+          {...rest}
+        />
+      )}
+    </>
+  )
+}
+
+Typeahead.propTypes = {
+  name: PropTypes.string.isRequired,
+  inputValue: PropTypes.string.isRequired,
+  value: PropTypes.string,
+  noOptionsMessage: PropTypes.func,
+  options: PropTypes.array,
+  styles: PropTypes.object,
+}
+
+Typeahead.defaultProps = {
+  value: '',
+  options: null,
+  styles: {},
+  noOptionsMessage: () => 'No options found',
+}
+
+Option.propTypes = {
+  inputValue: PropTypes.string.isRequired,
+  data: PropTypes.shape({
+    label: PropTypes.string,
+  }).isRequired,
+}
+
+export default Typeahead
--- a/src/typeahead/__stories__/Typeahead.stories.jsx
+++ b/src/typeahead/__stories__/Typeahead.stories.jsx
@@ -1,0 +1,210 @@
+/* eslint "react/destructuring-assignment": 0, react/prop-types: 0 */
+
+import React, { useState } from 'react'
+import { storiesOf } from '@storybook/react'
+import transformOptions from '../transformOptions'
+
+import Typeahead from '../Typeahead'
+import SmallTypeahead from '../SmallTypeahead'
+
+const options = [
+  { value: '1234', label: 'Chocolate', subLabel: 'mint' },
+  { value: '5678', label: 'Strawberry', subLabel: 'Cream' },
+  { value: '9876', label: 'Vanilla', subLabel: 'Vanilla' },
+]
+
+const asyncOptions = [
+  {
+    value: '379f390a-e083-4a2c-9cea-e3b9a08606a7',
+    label: 'Holly Collins',
+    subLabel: 'Heart of the South West LEP',
+  },
+  {
+    value: '8dcd2bb8-dc73-4a42-8655-4ae42d4d3c5a',
+    label: 'Bernard Harris-Patel',
+    subLabel: 'Welsh Government (Investment)',
+  },
+  {
+    value: 'a6f39399-5bf4-46cb-a686-826f73e9f0ca',
+    label: 'Dennis Kennedy',
+  },
+]
+
+storiesOf('Typeahead/Single select', module).add('Standard options', () =>
+  React.createElement(() => {
+    const [inputValue, setInputValue] = useState('')
+
+    const onInputChange = (val) => {
+      setInputValue(val)
+    }
+    return (
+      <div style={{ maxWidth: '400px' }}>
+        <Typeahead
+          onInputChange={onInputChange}
+          inputValue={inputValue}
+          isMulti={false}
+          closeMenuOnSelect={false}
+          name="test_1"
+          options={transformOptions(options)}
+          placeholder="Search..."
+        />
+      </div>
+    )
+  })
+)
+
+storiesOf('Typeahead/Single select', module).add(
+  'Options - pre-selected option',
+  () => {
+    return (
+      <div style={{ maxWidth: '400px' }}>
+        <Typeahead
+          isMulti={false}
+          closeMenuOnSelect={false}
+          label="Options - pre-selected option"
+          name="test_2"
+          options={options}
+          placeholder="Search..."
+          value={options[2].value}
+        />
+      </div>
+    )
+  }
+)
+
+storiesOf('Typeahead/Single select', module).add('Small', () => {
+  return (
+    <div style={{ maxWidth: '400px' }}>
+      <SmallTypeahead
+        isMulti={false}
+        closeMenuOnSelect={false}
+        name="test_2"
+        options={transformOptions(options)}
+        placeholder="Search..."
+      />
+    </div>
+  )
+})
+
+storiesOf('Typeahead/Multiple select', module)
+  .add('Standard options', () =>
+    React.createElement(() => {
+      const [inputValue, setInputValue] = useState('')
+
+      const onInputChange = (val) => {
+        setInputValue(val)
+      }
+      return (
+        <div style={{ maxWidth: '400px' }}>
+          <Typeahead
+            onInputChange={onInputChange}
+            inputValue={inputValue}
+            isMulti={true}
+            closeMenuOnSelect={false}
+            label="Standard options"
+            name="test_1"
+            options={transformOptions(options)}
+            placeholder="Search..."
+          />
+        </div>
+      )
+    })
+  )
+
+  .add('Options - pre-selected option', () =>
+    React.createElement(() => {
+      const [inputValue, setInputValue] = useState('')
+
+      const onInputChange = (val) => {
+        setInputValue(val)
+      }
+      return (
+        <div style={{ maxWidth: '400px' }}>
+          <Typeahead
+            onInputChange={onInputChange}
+            inputValue={inputValue}
+            isMulti={true}
+            closeMenuOnSelect={false}
+            name="test_1"
+            options={transformOptions(options)}
+            placeholder="Search..."
+            value={options[2].value}
+          />
+        </div>
+      )
+    })
+  )
+
+  .add('Async options', () =>
+    React.createElement(() => {
+      const [inputValue, setInputValue] = useState('')
+
+      const getOptions = () => {
+        return new Promise((resolve) => {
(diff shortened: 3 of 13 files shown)


ian-leggett:
> @ian-leggett For the async version, do you think that showing 'no advisors found' when a user first clicks in the type ahead could be misleading? Is it possible to show different text if they search and don't find something vs if they haven't searched anything yet?
> 
> Screenshot 2020-02-04 at 10 10 20

Good point but don't worry this is just an example of what you can use, you could remove this message if you like or pass any message you like.

src/typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,121 @@
...
+    components: {
+      Option: (props) => {
+        return <Option {...props} inputValue={inputValue} />
+      },
+    },
+  }
+
+  const filterOptions = ({ label: labelToFilter }) =>
+    labelToFilter.toLowerCase().includes(inputValue) ||
+    (labelToFilter && labelToFilter.toLowerCase().includes(inputValue))
```
web-bert: I think this could be clearer, I'm not sure what is trying to be accomplished  with the `labelToFilter &&` as it will have already blown up with the `labelToFilter.toLowerCase()` ? 

Maybe something like:
```JavaScript
function filterOptions({label}){
  const labelToFilter = label ? label.toLowerCase() : ''
  return labelToFilter.includes(inputValue)
}
```

src/typeahead/__stories__/Typeahead.stories.jsx
```diff
@@ -0,0 +1,186 @@
...
+  {
+    value: '379f390a-e083-4a2c-9cea-e3b9a08606a7',
+    label: 'Holly Collins',
+    subLabel: 'Heart of the South West LEP',
+  },
+  {
+    value: '379f390a-e083-4a2c-9cea-e3b9a08606a7',
+    label: 'Holly Collins',
+    subLabel: 'Heart of the South West LEP',
+  },
```
web-bert: Is this intentional to have the same option twice?

src/typeahead/Typeahead.jsx
ian-leggett: Yeah I agree, I have refactored this and its much clearer to read now. See what you think 👍

src/typeahead/__stories__/Typeahead.stories.jsx
ian-leggett: It was at first as I was testing against duplicates but I have removed this now as we don't need it.

src/typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,120 @@
...
+    styles: customStyles,
+    components: {
+      Option: (props) => {
+        return <Option {...props} inputValue={inputValue} />
+      },
+    },
+  }
+
+  const filterOptions = ({ label: labelToFilter }) =>
+    labelToFilter ? labelToFilter.toLowerCase().includes(inputValue) : ''
```
web-bert: should this return `false` instead of `''` as I assume it is going into an `array.filter` call?

ian-leggett: Good spot this should have been false thanks 👍

src/typeahead/Highlighter.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import reactStringReplace from 'react-string-replace'
+import styled from 'styled-components'
+
+/* istanbul ignore file */
+const Highlighter = ({ optionLabel, searchStr }) => {
+  const StyledSpan = styled('span')`
+    font-weight: 900;
+  `
```
peterhudec: This needs to go out of the `Highlighter` body, otherwise react will re-render it every time

src/typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const filterOptions = ({ label: labelToFilter }) =>
+    labelToFilter ? labelToFilter.toLowerCase().includes(inputValue) : false
+
+  return (
+    <>
+      <Label htmlFor={name}>
+        <StyledLabelText>{label}</StyledLabelText>
+        {hint && <StyledHintText>{hint}</StyledHintText>}
+      </Label>
```
peterhudec: I would not bake the label into this component. The label and hint don't take part in the typeahead logic and don't need to be here. If, then I would create a separate `LabeledTypeahead`:
```js
const LabeledTypeahead = ({label, hint, ...props}) =>
  <Label htmlFor={props.name}>
    <LabelText>{label}</LabelText>
    {hint && <HintText>{hint}</HintText>}
    <Typeahead {...props} />
  </Label>
```

src/typeahead/__tests__/Typeahead.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+      label="Standard options"
+      name="test_1"
+      options={transformOptions(selectOptions)}
+      placeholder="Search..."
+    />
+  )
+}
+
+describe('Typeahead', () => {
+  let wrapper
```
peterhudec: You don't need this `let`

src/typeahead/__tests__/Typeahead.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+  describe('<Select>', () => {
+    describe('when searching with the characters "choc"', () => {
+      wrapper = mount(<WrappedSelectTypeahead />)
```
peterhudec: prefix this with `const` and then you won't need the `let` above

src/typeahead/__tests__/Typeahead.test.jsx
```diff
@@ -0,0 +1,42 @@
...
+      const highlightedTextArray = wrapper.find('[data-test-id="test-1"]')
+      const highlightedText = highlightedTextArray.map((x) => x.text())[0]
+      test('should highlight the matching string', () => {
+        expect(highlightedText).toEqual('Choc')
+      })
+    })
+  })
+})
```
rafenden: We can add a bit more tests, here's a bare minimum:

```suggestion
import React from 'react'
import { mount } from 'enzyme'
import transformOptions from '../transformOptions'

import Typeahead from '../Typeahead'
import Highlighter from '../Highlighter'

const OPTIONS = [
  { value: '1234', label: 'Chocolate', subLabel: 'mint' },
  { value: '5678', label: 'Strawberry', subLabel: 'Cream' },
  { value: '9876', label: 'Vanilla', subLabel: 'Vanilla' },
]

const BASIC_PROPS = {
  name: 'test',
  options: transformOptions(OPTIONS),
  menuIsOpen: true,
}

describe('Typeahead', () => {
  test('highlights typed text', () => {
    const selectWrapper = mount(
      <Typeahead {...BASIC_PROPS} inputValue="choc" />
    )

    expect(selectWrapper.find(Highlighter)).toHaveLength(3)

    expect(selectWrapper.text()).toEqual(
      'Select...Chocolate - mintStrawberry - CreamVanilla - Vanilla'
    )

    expect(
      selectWrapper
        .find(Highlighter)
        .at(0)
        .find('span')
        .text()
    ).toEqual('Choc')
  })
})

```

src/typeahead/transformOptions.js
```diff
@@ -0,0 +1,16 @@
+/* istanbul ignore file */
+const transformOptions = (options) => {
+  return options.map(({ value, label, subLabel }) => {
+    return subLabel
+      ? {
+          value,
+          label: `${label} - ${subLabel}`,
```
rafenden: Why not to simply use `label` only? What's the special use for `subLabel`?

src/typeahead/Typeahead.jsx
ian-leggett: So I would use this outside but it doesn't play nicely as if I wrap it with a label and hint etc the event handler gets eaten up inside the typeahead (bubbling issue). I see this similar to the "Field input text" component we have where you do get the label and hint. I can't think of a scenario when we would want this typeahead without a label.

src/typeahead/__tests__/Typeahead.test.jsx
ian-leggett: Changed the tests :-)

src/typeahead/__tests__/Typeahead.test.jsx
ian-leggett: Thanks for this, I have replaced your tests over mine but I can't think of how I can do more tests. Ideally we need it to simulate the typing of text in the input but it just won't play nicely. The above is much cleaner so I used it but it doesn't give me much more than what I had apart from asserting 3 options which we already know we have. It would be nice to see that when we search for "Choc" we have one option left and its characters "Choc" are highlighted. I would also like to test the async version but I can't type in the text field to fire the onChange handler :-(

src/typeahead/transformOptions.js
ian-leggett: So if you wanted to add the team beside the adviser such as `Rafal - DIT` you will need to concat the label and subLabel as we did in the Vue version of the typeahead. If there is no subLabel(team) we just render the label.

src/typeahead/__tests__/Typeahead.test.jsx
rafenden: Looks like `react-select` doesn't test well https://stackoverflow.com/questions/41991077/testing-react-select-component

src/typeahead/transformOptions.js
rafenden: Well, a good reason to have them separately would be a consistent separator.

src/typeahead/Highlighter.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import reactStringReplace from 'react-string-replace'
+import styled from 'styled-components'
+import { FONT_WEIGHTS } from '@govuk-react/constants'
+
+/* istanbul ignore file */
```
rafenden: Please remove.

src/typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,123 @@
+/* istanbul ignore file */
```
rafenden: Please remove.

src/typeahead/__tests__/Typeahead.test.jsx
```diff
@@ -0,0 +1,39 @@
+/* istanbul ignore file */
```
rafenden: No need for this in the test itself.

src/typeahead/transformOptions.js
```diff
@@ -0,0 +1,16 @@
+/* istanbul ignore file */
```
rafenden: Please remove.

src/typeahead/styles.js
```diff
@@ -0,0 +1,82 @@
+/* istanbul ignore file */
```
rafenden: Please remove.

src/typeahead/__tests__/Typeahead.test.jsx
rafenden: Although, you could add tests for the custom props, e.g. `reduceSize`

src/typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,123 @@
...
+  ...rest
+}) => {
+  const StyledLabelText = styled(LabelText)`
+    font-weight: ${reduceSize ? FONT_WEIGHTS.bold : FONT_WEIGHTS.regular};
+    ${MEDIA_QUERIES.TABLET} {
+      font-size: ${reduceSize ? FONT_SIZE.SIZE_16 : FONT_SIZE.SIZE_19};
+    },
+  `
+  const StyledHintText = styled(HintText)`
+    margin-bottom: ${SPACING.SCALE_2};
```
peterhudec: These two need to go out of the component otherwise React would always re-render them

src/typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const Typeahead = ({
+  name,
+  value,
+  label,
+  hint,
+  options,
+  inputValue,
+  reduceSize,
```
peterhudec: I would remove the `reduceSize`. This is against the [Open–closed SOLID](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle) principle, which says:
> software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification

You baked in a particular style variant instead of allowing it to be tweaked from the outside. If anyone would be in need of another variant, they would need to implement it here. I would do this exactly as the underlying library i.e. would allow to pass the whole `styles` with the styles yuo defined for our `Typeahead` as its default value. Then you could implement a `SmallTypeahead` by just wrapping `Typeahead` and overriding its `styles`.

src/typeahead/__tests__/Typeahead.test.jsx
ian-leggett: After talking with Peter we have removed the label and hint out of this component as it made sense not to bake this in so these props are removed now.

src/typeahead/Typeahead.jsx
ian-leggett: Makes sense, I have now created a `<SmallTypeahead>` 👍

src/typeahead/Highlighter.jsx
```diff
@@ -0,0 +1,27 @@
...
+const Highlighter = ({ optionLabel, searchStr }) => {
+  if (!optionLabel) {
+    return null
+  }
+  return reactStringReplace(optionLabel, searchStr, (matchedTxt, i) => (
+    <StyledSpan key={i} data-test-id={`test-${i}`}>
+      {matchedTxt}
+    </StyledSpan>
+  ))
+}
```
peterhudec: This looks more complicated than it actually is. I would reverse the condition:
```js
const Highlighter = ({ optionLabel = null, searchStr }) =>
  optionLabel && reactStringReplace(optionLabel, searchStr, (matchedTxt, i) =>
    <StyledSpan key={i} data-test-id={`test-${i}`}>
      {matchedTxt}
    </StyledSpan>
  )
```

ian-leggett: Good point, all done 👍

src/typeahead/Typeahead.jsx
```diff
@@ -0,0 +1,91 @@
...
+    </components.Option>
+  )
+}
+
+const Typeahead = ({
+  name,
+  value,
+  options,
+  inputValue,
+  customStyles,
```
peterhudec: I think this should be `styles` and not `customStyles` to preserve the `react-select` API.

src/typeahead/__tests__/Typeahead.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+]
+const BASIC_PROPS = {
+  name: 'test',
+  label: 'Advisers',
+  hint: 'Some hint',
+  options: transformOptions(OPTIONS),
+  menuIsOpen: true,
+}
+describe('Typeahead', () => {
+  let selectWrapper
```
peterhudec: You don't need to define a `let` here. Just define a `const wrapper = mount(...)` in each tests.

src/typeahead/__tests__/Typeahead.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+  test('renders options', () => {
+    selectWrapper = mount(<Typeahead {...BASIC_PROPS} inputValue="" />)
```
peterhudec: Prefix this with `const` and you wouldn't need the `let` above.

src/typeahead/__tests__/Typeahead.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+    expect(selectWrapper.find(components.Option)).toHaveLength(3)
+    expect(selectWrapper.find('MenuList').text()).toEqual(
+      'Chocolate - mintStrawberry - CreamVanilla - Vanilla'
+    )
+  })
+  test('highlights typed text', () => {
+    selectWrapper = mount(<Typeahead {...BASIC_PROPS} inputValue="choc" />)
```
peterhudec: Prefix this with `const` and you wouldn't need the `let` above.

src/typeahead/__tests__/Typeahead.test.jsx
peterhudec: I can still see the `let` in there

PR #254, opened 2020-02-07
fix: Update the content within FieldDnbCompany

rafenden:
Content updates to the D&B company search.

### DIFF
--- a/src/forms/elements/__tests__/FieldDnbCompany.test.jsx
+++ b/src/forms/elements/__tests__/FieldDnbCompany.test.jsx
@@ -225,7 +225,7 @@ describe('FieldDnbCompany', () => {
     })
 
     test('should render cannot find actions', () => {
-      expect(wrapper.text()).toContain('Try improving your search by')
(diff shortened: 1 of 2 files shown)


src/forms/elements/__tests__/FieldDnbCompany.test.jsx
```diff
@@ -295,7 +295,7 @@ describe('FieldDnbCompany', () => {
     })
   })
 
-  describe('when "I still cannot find the company" link is clicked', () => {
+  describe('when "I still cant find what Im looking for" link is clicked', () => {
```
ian-leggett: Optional and very small - "I still can't find what I'm looking for"

rafenden: Apostrophes were omitted intentionally, it's only a test description and I wanted to avoid escaping them.

rafenden: I used template literals in the end.

PR #255, opened 2020-02-10
ci: Enable Docker Layer Caching on CircleCI to speed-up tests

rafenden:
Enable [caching of Docker layers](https://circleci.com/docs/2.0/docker-layer-caching/) which will speed-up our tests.


rafenden:
Closing as this is not applicable to our build setup.

> Note: DLC has no effect on Docker images used as build containers. That is, containers that are used to run your jobs are specified with the image key when using the docker ...

https://circleci.com/docs/2.0/docker-layer-caching/

PR #260, opened 2020-02-18
refactor: Add options to currency formatter

paulgain:
Pass an options object to our currency functions allowing consumers to format currency based on their requirements.

A list of options can be found here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat

### DIFF
--- a/src/utils/NumberUtils.js
+++ b/src/utils/NumberUtils.js
@@ -11,7 +11,7 @@ export default class NumberUtils {
     return formatter.format(number)
   }
 
-  static currencyGBP(number) {
+  static currencyGBP(number, options = { maximumSignificantDigits: 21 }) {
     if (!number && number !== 0) {
       return null
     }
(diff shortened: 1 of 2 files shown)


src/utils/NumberUtils.js
```diff
@@ -11,29 +11,37 @@ export default class NumberUtils {
...
     return formatter.format(number)
   }
 
-  static currencyUSD(number) {
+  static currencyUSD(number, options) {
     if (!number && number !== 0) {
       return null
     }
 
+    const opts = options || {
```
rafenden: This could be passed as a default value for func argument.

src/utils/NumberUtils.js
```diff
@@ -11,29 +11,37 @@ export default class NumberUtils {
     return formatter.format(number)
   }
 
-  static currencyGBP(number) {
+  static currencyGBP(number, options) {
     if (!number && number !== 0) {
       return null
     }
 
+    const opts = options || {
```
rafenden: This could be passed as a default value for func argument.

src/utils/NumberUtils.js
paulgain: 👍

src/utils/NumberUtils.js
paulgain: 👍

PR #261, opened 2020-02-24
Add referrals to activity feed

ian-leggett:
## Description
Trello - https://trello.com/c/L0XaqC5U/1080-fe-show-referrals-on-activity-feed-5
We need to add the completed and outstanding referrals to the Activity feed.

Screenshot 2020-02-24 at 15 22 28

### DIFF
--- a/src/activity-feed/activities/Referral.jsx
+++ b/src/activity-feed/activities/Referral.jsx
@@ -1,0 +1,67 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Link from '@govuk-react/link'
+import DateUtils from '../../utils/DateUtils'
+import { Card, CardHeader, CardTable } from './card'
+import { ACTIVITY_TYPE } from '../constants'
+import CardUtils from './card/CardUtils'
+import ReferralUtils from './ReferralUtils'
+
+export default class Referral extends React.PureComponent {
+  static propTypes = {
+    activity: PropTypes.object.isRequired,
+  }
+
+  static canRender(activity) {
+    return CardUtils.canRenderByTypes(activity, ACTIVITY_TYPE.Referral)
+  }
+
+  render() {
+    const { activity } = this.props
+    const {
+      id,
+      subject,
+      startTime,
+      sender,
+      recipient,
+      completedOn,
+    } = CardUtils.transformReferral(activity)
+    const url = `/referral/${id.split(':')[2]}`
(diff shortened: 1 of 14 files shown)


src/activity-feed/activities/Referral.jsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Link from '@govuk-react/link'
+import DateUtils from '../../utils/DateUtils'
+import { Card, CardHeader, CardTable } from './card'
+import { ACTIVITY_TYPE } from '../constants'
+import CardUtils from './card/CardUtils'
+import ReferralUtils from './ReferralUtils'
+
+export default class Referral extends React.PureComponent {
```
debitan: I see what you did hear, matching the coding style to that of activity feed. However, as this component can be a functional component without any ill effects on the rest of activity, I think we should refactor it to be a functional component in our now agreed style.

src/activity-feed/activities/ReferralUtils.js
```diff
@@ -0,0 +1,9 @@
+import { get } from 'lodash'
+import { BADGES } from '../constants'
+
+export default class ReferralUtils {
```
debitan: same here. I would do a simple functional component.

src/activity-feed/activities/card/CardUtils.jsx
```diff
@@ -65,6 +65,35 @@ const getAdviser = (activity) => {
...
+      email: get(sender, 'dit:emailAddress'),
+      team: get(sender, 'dit:team[name]'),
+    },
+    recipient: {
+      name: get(recipient, 'name'),
+      email: get(recipient, 'dit:emailAddress'),
+      team: get(recipient, 'dit:team[name]'),
+    },
+    completedOn: get(activity, 'object.dit:completedOn'),
+  }
```
debitan: OPTIONAL: using `get` is following the coding pattern already established in this file. However, we could use `?.` optional chaining from ES2020 to reduce our reliance of lodash.

src/activity-feed/activities/Referral.jsx
ian-leggett: I started to do some refactoring but because its all classed based style this is a bit of a pain to do just part of it. Ideally we need to refactor the whole thing to be functional.

PR #262, opened 2020-02-25
refactor: Lift forms state to enable use of Redux and similar libraries

rafenden:
Make the internal forms state optional.


src/forms/elements/Form.jsx
```diff
@@ -4,75 +4,34 @@ import LoadingBox from '@govuk-react/loading-box'
...
-    </form>
-  )
-}
-
-FormWrapper.propTypes = {
-  children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
-}
-
-FormWrapper.defaultProps = {
+  isLoading: false,
```
debitan: we discussed this. As the `Task` component handles all of the fetching and posting of data, and in turn the loading state, this might not be necessary. But as it is defaulted to false and therefore optional we can leave it in. It gives us the option of using the `Task` component to handle loading or in another situation handle it within the component itself.

mcbhenwood:
It would be useful for me to understand how this proposed change would affect the discussion about the divergence of implementations of forms in DH FE. It's not clear to me how the two approaches would differ following this change. Can you write a brief explanation @rafenden or even maybe hop on a call with me about it this afternoon?

peterhudec:
> I don't really see how this change makes it pluggable to redux. What I was expecting from this PR is a story where the state would be managed with a an external reducer, but the only change I see here is the local state being _lifted_ from the `Form` to the `FormStateProvider`, but its still the internal state managed in the [`useForm`](https://github.com/uktrade/data-hub-components/blob/master/src/forms/hooks/useForm.js#L27) hook. I'm afraid nothig at all has changed from the redux perspective.
> 
> I could still imagine to make this work though. You would need to move the whole `useForm` logic to a reducer and make the whole `Form` completely stateless. You could however keep the context connection between the form and its various fields.

@rafenden Sorry, the name of `FormStateProvider` confused me and now I can see how the props are spread in the story to the `Form` and it indeed seems that the `Form` is now stateless.

src/forms/hooks/useForm.js
```diff
@@ -27,7 +27,7 @@ function useBeforeUnload(enabled, onExit) {
 function useForm({
   initialValues = {},
   initialStep = 0,
-  onSubmit = null,
+  onSubmit,
```
mforner13: Just to make sure I understand, will this make it possible to connect to an action/reducer on submit?

src/forms/elements/FormStateful.jsx
```diff
@@ -0,0 +1,44 @@
...
+FormStateful.defaultProps = {
+  initialValues: {},
+  initialStep: 0,
+  onSubmit: undefined,
+  scrollToTop: true,
+  onExit: undefined,
+  formAttributes: {},
+}
+
+export default FormStateful
```
mforner13: Is the plan to then replace all existing uses of the Form component on DH with FormStateful?

rafenden: Yes, only until it's replaced with Redux which will use the `Form` component directly.

src/forms/hooks/useForm.js
rafenden: This will be possible by using the`Form` component directly, here's an example implementation (the current form components support steps):
```
function submitHandler = (formState) => {
  if (!formState.isLastStep()) {
    dispatch({
      type: FORM_STEP_FORWARD,
      formState,
    })
  } else {
    dispatch({
      type: SEND_REFERRAL_FORM__SUBMIT,
      formState,
    })
  }
}

function setFieldValue = (fieldName, formState) => 
  dispatch({
    type: FORM_FIELD_CHANGE,
    fieldName,
    formState,
  })

<Form setFieldValue={setFieldValue} goForward={submitHandler}>
  <Step name="details">
    <FieldTypeahead
      name="Advisor"
    />
    
    <FieldInput
      name="subject"
      label="Subject"
    />
  </Step>
  <Step name="confirmation">
    <SummaryTable caption="Check referral details">
      ...
    </SummaryTable>
  </Step>
</Form>
```

src/forms/elements/Form.jsx
peterhudec: You can actually wrap this in a `Task`:
```js
<Task>
  {task => <Form isLoading={task('task-name', 'task-id').progress} />}
</Task>
```

rafenden: The `Loader` was now removed from the `Form` component.

PR #266, opened 2020-03-04
fix: use index for CollectionItem key

debitan:
Use the index for the CollectionItem key as the current use of heading and url result in clashing keys in the country specific exportHistory implementation of this component in the frontend.

There isn't a unique id provided for these components, so the next best option is to use the index. This isn't normally advised and hence the `/* eslint-disable react/no-array-index-key */`

This is detailed in the docs here - https://reactjs.org/docs/lists-and-keys.html#keys

I've also added a test for this.

### DIFF
--- a/src/collection-list/CollectionList.jsx
+++ b/src/collection-list/CollectionList.jsx
@@ -1,3 +1,6 @@
+/* eslint-disable react/no-array-index-key */
+// this is because there isn't necessarily a unique id to use as the key
+
 import React from 'react'
 import PropTypes from 'prop-types'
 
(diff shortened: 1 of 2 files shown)


mforner13:
> Not sure if this is possible, but could you prefix the key with an 'export' differentiator if it's used in the export history?

We talked, realised the key clashing is happening on the exports history tab, not across the activity feed and the exports history.

web-bert:
From the Keys doc above it states 

> If you choose not to assign an explicit key to list items then React will default to using indexes as keys.

So can we just remove the key instead?

debitan:
> From the Keys doc above it states
> 
> > If you choose not to assign an explicit key to list items then React will default to using indexes as keys.
> 
> So can we just remove the key instead?

@web-bert We could, but then we'd get errors in the console.

PR #268, opened 2020-03-05
fix: render no pagination when there are 0 results

debitan:
My previous bug fix introduced this small bug. Before when you had 0 results the pagination would render. Now when the `totalPages` === 0 the pagination component won't render.

### DIFF
--- a/src/collection-list/__tests__/CollectionList.test.jsx
+++ b/src/collection-list/__tests__/CollectionList.test.jsx
@@ -7,7 +7,7 @@ import CollectionHeader from '../CollectionHeader'
 import CollectionItem from '../CollectionItem'
 import CollectionDownload from '../CollectionDownload'
 
-describe('CollectionItem', () => {
+describe('CollectionList', () => {
   let wrapper
   const getPageUrl = (page) => `#page-${page}`
   const onPageClick = () => {}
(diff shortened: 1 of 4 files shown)


src/pagination/Pagination.jsx
```diff
@@ -74,7 +74,7 @@ const StyledPagesTruncation = styled('span')`
 function Pagination({ totalPages, activePage, onPageClick, getPageUrl }) {
   const visiblePieces = computeVisiblePieces(totalPages, activePage)
 
-  if (totalPages === 1) {
+  if (totalPages === 0 || totalPages === 1) {
```
rafenden: Optional: could be slightly simplified
```suggestion
  if (totalPages < 2) {
```

src/collection-list/__tests__/CollectionList.test.jsx
```diff
@@ -89,4 +89,14 @@ describe('CollectionItem', () => {
       expect(wrapper.find(Pagination).prop('getPageUrl')).not.toBeNull()
     })
   })
+
+  describe('when no items are passed', () => {
+    beforeAll(() => {
+      wrapper = mount(<CollectionList />)
+    })
+
+    test('should no render pagination', () => {
```
web-bert: missing the t for not

PR #275, opened 2020-03-11
Fix referral urls in activity feed

ian-leggett:
## Description
Small update to fix the links of referrals in the activity feed.


src/activity-feed/activities/card/CardUtils.jsx
```diff
@@ -77,6 +77,7 @@ const transformReferral = (activity) => {
 
   return {
```
rafenden: Please move the `transformReferral()` function to the `Referral.jsx` or `ReferralUtils.js` as it's specific to Referral activity.

ian-leggett: Good point, just moved this to `ReferralUtils.js` 👍

PR #283, opened 2020-03-20
feat: Add date component

paulgain:
# Date component
A new date component that allows default and custom validation of dates.

### DIFF
--- a/src/forms/elements/FieldDate.jsx
+++ b/src/forms/elements/FieldDate.jsx
@@ -1,0 +1,157 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import moment from 'moment'
+import { castArray } from 'lodash'
+import styled from 'styled-components'
+import { ERROR_COLOUR } from 'govuk-colours'
+import ErrorText from '@govuk-react/error-text'
+import Label from '@govuk-react/label'
+import Input from '@govuk-react/input'
+import LabelText from '@govuk-react/label-text'
+import {
+  BORDER_WIDTH_FORM_ELEMENT_ERROR,
+  SPACING,
+} from '@govuk-react/constants'
+
+import FieldWrapper from './FieldWrapper'
+import useField from '../hooks/useField'
+import { useFormContext } from '../..'
+
+const VALIDATION_DATE_FORMAT = 'YYYY-MM-DD'
+const DAY = 'day'
+const MONTH = 'month'
+const YEAR = 'year'
+
+const StyledInputWrapper = styled('div')`
+  ${(props) =>
+    props.error &&
+    `
+    border-left: ${BORDER_WIDTH_FORM_ELEMENT_ERROR} solid ${ERROR_COLOUR};
+    margin-right: ${SPACING.SCALE_3};
+    padding-left: ${SPACING.SCALE_2};
+  `}
+`
+
+const StyledLabel = styled(Label)(
+  {
+    marginRight: '20px',
+    marginBottom: 0,
+  },
+  ({ year }) => ({
+    width: year ? '70px' : '50px',
+  })
+)
+
+const StyledList = styled('div')({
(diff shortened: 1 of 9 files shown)


src/forms/elements/__stories__/FieldDate.stories.jsx
```diff
@@ -0,0 +1,57 @@
...
+addDecorator(withKnobs)
+
+const ERRORS = {
+  DAY: 'Enter a valid Epoch day',
+  MONTH: 'Enter a valid Epoch month',
+  YEAR: 'Enter a valid Epoch year',
+}
+
+storiesOf('Forms', module)
+  .add('FieldDate - default validation', () => (
```
rafenden: I suggest merging this two stories into one.

src/forms/elements/FieldDate.jsx
```diff
@@ -0,0 +1,154 @@
...
+import {
+  BORDER_WIDTH_FORM_ELEMENT_ERROR,
+  SPACING,
+} from '@govuk-react/constants'
+
+import FieldWrapper from './FieldWrapper'
+import useField from '../hooks/useField'
+import { useFormContext } from '../..'
+
+const DATE_FORMAT = 'YYYY-MM-DD'
```
rafenden: I suggest we call it `VALIDATION_DATE_FORMAT` so it's clear that this is used only for validation.

src/forms/elements/FieldDate.jsx
```diff
@@ -0,0 +1,154 @@
...
+  return (
+    <FieldWrapper {...{ name, label, hint, error }}>
+      <StyledInputWrapper error={error}>
+        {error && <ErrorText>{error}</ErrorText>}
+        <StyledList>
+          <StyledLabel>
+            <LabelText>{labels.day}</LabelText>
+            <Input
+              id={`${name}-day`}
+              name={`${name}-day`}
```
rafenden: I gave some though on that recently and perhaps we should use dots here, e.g. `address.postcode` so in the future we could convert (on submission) form values from flat to complex object:
```
{
  values: {
    address.postcode: "SE1 123"
  }
}
```
becomes
```
{
  values: {
    address: {
       postcode: "SE1 123"
    }
  }
}
```
which will spare us some transformations

src/forms/elements/FieldDate.jsx
```diff
@@ -0,0 +1,154 @@
...
+          </StyledLabel>
+        </StyledList>
+      </StyledInputWrapper>
+    </FieldWrapper>
+  )
+}
+
+FieldDate.propTypes = {
+  name: PropTypes.string.isRequired,
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
rafenden: We can just use `PropTypes.node`

src/forms/elements/__stories__/FieldDate.stories.jsx
paulgain: I purposely split them so it's clear to developers you are given default date validation out of the box, meaning, you don't have to roll your own unless the default implementation doesn't fit your requirements.

src/forms/elements/FieldDate.jsx
paulgain: 👍

src/forms/elements/FieldDate.jsx
paulgain: 👍

src/forms/elements/FieldDate.jsx
paulgain: 👍

src/forms/elements/__stories__/FieldDate.stories.jsx
rafenden: Cool, it makes sense.

PR #288, opened 2020-03-23
fix: Export FieldDate component

paulgain:
Export the new component and remove a cyclic dependency.


mcbhenwood:
Ignore the above changes, was just testing something.

PR #289, opened 2020-03-25
Align the table headers in all activity feed cards

ian-leggett:
## Description
The cards table content(headers) in the activity feed are all miss-aligned. As the structure slightly differs between a referral and interaction I have now set a minimum width for both referral and interactions so they align correctly.

## Before
Screenshot 2020-03-25 at 10 01 29

## After
Screenshot 2020-03-25 at 13 46 14

### DIFF
--- a/src/activity-feed/activities/Referral.jsx
+++ b/src/activity-feed/activities/Referral.jsx
@@ -50,6 +50,7 @@ export default class Referral extends React.PureComponent {
         />
 
         <CardTable
+          isWideCard={true}
           rows={[
             { header: 'Sending adviser', content: AdviserDetails(sender) },
             {
--- a/src/activity-feed/activities/Referral.jsx
+++ b/src/activity-feed/activities/Referral.jsx
@@ -50,7 +50,7 @@ export default class Referral extends React.PureComponent {
         />
 
         <CardTable
-          isWideCard={true}
+          isNotWrappedInDetails={true}
           rows={[
             { header: 'Sending adviser', content: AdviserDetails(sender) },
             {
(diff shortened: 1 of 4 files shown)


src/activity-feed/activities/card/CardDetailsList.jsx
```diff
@@ -1,14 +1,18 @@
...
-const StyledUList = styled('ul')`
-  margin: 0;
-  padding-left: 0;
+const StyledListItem = styled(ListItem)`
+  ${MEDIA_QUERIES.TABLET} {
+    font-size: 16px;
+  }
 `
 
 export default class CardDetailsList extends React.PureComponent {
```
debitan: OPTIONAL: this looks an easy one to convert to a functional component if you fancied it

src/activity-feed/activities/Referral.jsx
```diff
@@ -50,6 +50,7 @@ export default class Referral extends React.PureComponent {
         />
 
         <CardTable
+          isReferral={true}
```
debitan: is this possibly too specific? I know that this component as a whole is too specific for a shared component, but we could start here and rename this prop to something that reflects the change in style? Not sure exactly what that could be, maybe `isWideCard` or `noBottomMargin` card?

src/activity-feed/activities/card/CardDetailsList.jsx
ian-leggett: Agreed, its worth tidying this component up as we go.

src/activity-feed/activities/Referral.jsx
ian-leggett: I couldn't think of a better name to be fair so I used `isWideCard` I think this works thanks

mforner13: @ian-leggett How would a user of the Card component know whether they should have isWideCard true or not? Could you name the prop after which bit of the referral card affects the alignment?

ian-leggett: Good point, after our chat I have changed the prop to `isNotWrappedInDetails` 👍

PR #294, opened 2020-04-17
Move styled-components to peer dependencies

peterhudec:
Having `styled-components` listed in `dependencies` results in having two `styled-components` instances installed in `data-hub-frontend` which also has it as a dependency:

Screen Shot 2020-04-17 at 12 35 41 PM

This then results in this error:
Screen Shot 2020-04-17 at 1 08 34 PM

This PR moves `styled-components` to [`peerDependencies`](https://docs.npmjs.com/files/package.json#peerdependencies) which should hopefully solve the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,12 +78,12 @@
     "react-markdown": "^4.0.8",
     "react-select": "^3.0.8",
     "react-string-replace": "^0.4.4",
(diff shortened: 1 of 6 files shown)


rafenden:
Fixes https://github.com/styled-components/styled-components/issues/1032

PR #295, opened 2020-04-27
feat: Add custom error message to FieldTypeahead

web-bert:
To enable showing a dynamic validation error we need more than just the `validate` prop, to give ultimate control the the parent add a simple `errorMessage` prop which will put the wrapper and typeahead in an error state and render the error message -  this should play nicely with the existing validation

### DIFF
--- a/src/forms/elements/__stories__/FieldTypeahead.stories.jsx
+++ b/src/forms/elements/__stories__/FieldTypeahead.stories.jsx
@@ -34,7 +34,7 @@ storiesOf('Forms', module).add('FieldTypeahead', () => (
         <FieldTypeahead
           label="Typeahead - sync single value"
           hint="Some hint"
-          name="sync_single"
+          name="sync_single_1"
           required="Chose value"
           options={options}
         />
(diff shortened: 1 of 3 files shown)


src/forms/elements/__stories__/FieldTypeahead.stories.jsx
```diff
@@ -69,6 +69,14 @@ storiesOf('Forms', module).add('FieldTypeahead', () => (
           loadOptions={getOptions}
           isMulti={true}
         />
+        <FieldTypeahead
+          label="Typeahead - sync with error message"
+          hint="Some hint"
+          name="sync_single_error"
+          required="Chose value"
+          errorMessage="My custom error message"
```
paulgain, line 77: For most, if not all, form components `required` is used for defining custom error messages. Providing a second prop `errorMessage ` for another error message is not ideal as it will lead to confusion as to what prop should be set. I'm not sure what you trying to achieve?

debitan: That's a good point @paulgain . Could this be solved by renaming `errorMessage` to something like `dynamicValidationErrorMessage`. That should make the distinction clearer.

paulgain: Up until now @debitan we have only validated on `onSubmit` (inline with GDS guidelines), however, we now have a valid use case (there will be more in the future) where we need to validate certain fields `onChange`. How we go about that is a bigger discussion and something @web-bert  will bring up in the Dev Huddle tomorrow.

PR #299, opened 2020-05-14
fix: Allow setFieldValue to accept empty strings

LaurenQurashi:
This PR relates to another PR in the DH front end: https://github.com/uktrade/data-hub-frontend/pull/2672

This PR fixes a bug with the useForm hook, which initially omitted a form field that had been passed an empty string. This would cause the unwanted behaviour of not updating an optional field to an empty string, if a value had been previously submitted or stored on the API. 

The fix is to simply remove the code that would check if the string was empty and omit it. 

Tests have been amended to reflect that we now want to accept an empty string as a value.

### DIFF
--- a/src/forms/hooks/__tests__/useForm.test.jsx
+++ b/src/forms/hooks/__tests__/useForm.test.jsx
@@ -278,12 +278,12 @@ describe('useForm', () => {
       })
     })
 
-    test('should unset the field value', () => {
-      expect(formState.values).toEqual({})
+    test('should set the field value', () => {
+      expect(formState.values).toEqual({ testField: '' })
     })
 
-    test('should set "isDirty" to false', () => {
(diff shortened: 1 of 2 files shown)


paulgain:
@LaurenQurashi

Previously
```javascript
// Form values
const values: { a: '', b: 'foo' }
// Strip out any object key when the value is an empty string.
const values: { b: 'foo' }
```

Afterwards
```javascript
// Form values
const values: { a: '', b: 'foo' }
// Keep all keys that are an empty string
const values: { a: '', b: 'foo' }
```

When the `fieldValue` is an empty `string` we want to keep that in the form values object. When a user clears a previously saved field within a form the subsequent POST body will now contain a key where its value is an empty string, due to the the existence of this key the API will be updated accordingly.

PR #300, opened 2020-05-19
fix: Search DnB companies - only set the postcode if it's defined

paulgain:
Steps to reproduce (Add company):
1. Go to dev `/companies/create`
2. Select UK and continue
3. Add a company name such as Lambda and search (yields results)
4. Add any postcode and search again (yields zero results)
5. Clear the postcode field and search again (error HTTP 400)

The problem lies in the existence of an empty postcode field in the body of the POST. The D&B API takes both the company name and the postcode and applies AND Boolean Logic when querying their database. The API errors whenever the postcode is an empty string.

To fix the issue we need to ensure the `postal_code` field is not included in the body of the POST when it's an empty string.

### DIFF
--- a/src/forms/elements/FieldDnbCompany.jsx
+++ b/src/forms/elements/FieldDnbCompany.jsx
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react'
 import PropTypes from 'prop-types'
-import { isEmpty } from 'lodash'
+import { isEmpty, omitBy } from 'lodash'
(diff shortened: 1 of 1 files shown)


paulgain:
> You ❤️lodash! But this is a great use of it. To write that natively would be really messy.

@debitan too right :smiley: `omit`, `omitBy`, `pick` and `pickyBy` save on unnecessary code and corresponding tests.

PR #304, opened 2020-06-16
Add a FieldAddress onSelect callback and preselect a FieldSelect option

paulgain:


src/forms/elements/__tests__/FieldSelect.test.jsx
```diff
@@ -113,6 +113,31 @@ describe('FieldSelect', () => {
     })
   })
 
+  describe('when the field specifies a preselected option', () => {
+    let options
+
+    beforeAll(() => {
```
ian-leggett, line 119: Do you need this lifecycle hook here as you only have one test in this suite

src/forms/elements/__tests__/FieldAddress.test.jsx
```diff
@@ -165,7 +165,9 @@ describe('FieldAddress', () => {
   })
 
   describe('when selecting an address', () => {
+    const onSelectUKAddressSpy = jest.fn()
     let axiosMock
+
     beforeAll(async () => {
```
ian-leggett, line 171: Do you need this lifecycle hook here as you only have one test in this suite

src/forms/elements/__tests__/FieldSelect.test.jsx
paulgain: Jest doesn't have a `before`: https://jestjs.io/docs/en/api

PR #305, opened 2020-06-17
fix: correctly select a <select> option

paulgain:
Use the `value` instead of `defaultValue`
https://reactjs.org/docs/forms.html#the-select-tag

### DIFF
--- a/src/forms/elements/__tests__/FieldSelect.test.jsx
+++ b/src/forms/elements/__tests__/FieldSelect.test.jsx
@@ -113,31 +113,6 @@ describe('FieldSelect', () => {
     })
   })
 
-  describe('when the field specifies a preselected option', () => {
-    let options
-
-    beforeAll(() => {
-      wrapper = mount(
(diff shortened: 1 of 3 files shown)


src/forms/elements/__tests__/FieldSelect.test.jsx
```diff
@@ -200,10 +175,12 @@ describe('FieldSelect', () => {
         .simulate('change', { target: { value: 'testOptionValue2' } })
     })
 
-    test('should update field value', () => {
-      expect(wrapper.find('select').prop('defaultValue')).toEqual(
-        'testOptionValue2'
-      )
+    test('should update field the value', () => {
```
ian-leggett, line 178: Should this be "should update field value"?

paulgain: Ha, you know it should, thanks @ian-leggett.

```javascript
test('should make more sense', () => { ... }
```

PR #310, opened 2020-06-24
fix: select option fix

paulgain:
It appears when using React's [recommended](https://reactjs.org/docs/forms.html#the-select-tag) approach to select an option combined with a [govuk-react](https://github.com/govuk-react/govuk-react) [select](https://github.com/govuk-react/govuk-react/tree/master/components/select) we don't get a `selected` attribute on the currently selected option which is causing some FE tests to fail.

### DIFF
--- a/src/forms/elements/__tests__/FieldSelect.test.jsx
+++ b/src/forms/elements/__tests__/FieldSelect.test.jsx
@@ -175,12 +175,19 @@ describe('FieldSelect', () => {
         .simulate('change', { target: { value: 'testOptionValue2' } })
     })
 
-    test('should update the field value', () => {
-      expect(wrapper.find('select').prop('value')).toEqual('testOptionValue2')
+    test('should update the defaultValue', () => {
+      expect(wrapper.find('select').prop('defaultValue')).toEqual(
+        'testOptionValue2'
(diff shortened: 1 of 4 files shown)


src/forms/elements/__tests__/FieldSelect.test.jsx
```diff
@@ -175,12 +175,10 @@ describe('FieldSelect', () => {
         .simulate('change', { target: { value: 'testOptionValue2' } })
     })
 
-    test('should update the field value', () => {
-      expect(wrapper.find('select').prop('value')).toEqual('testOptionValue2')
-    })
-
-    test('should set the readonly flag', () => {
-      expect(wrapper.find('select').prop('readOnly')).toEqual(true)
+    test('should update the defaultValue', () => {
```
web-bert, line 178: There should be a test to confirm that the `<option>` gets the `selected` attribute added so this doesn't happen again, as that is the root cause of the problem here: https://github.com/uktrade/data-hub-frontend/blob/master/test/functional/cypress/support/assertions.js#L129

paulgain: 👍

PR #317, opened 2020-07-13
Update npm packages to fix security alerts

ian-leggett:
## Description
Dependabot is flagging up security vulnerability alerts with two dependancies: "acorn" and "kind-of". 

## Solution
As "acorn" lives in Jest I decided to get Jest up to date so I have now bumped this up two major versions. As for "kind-of" I had some trouble resolving this with yarn so to fall in sync with Data Hub FE I have removed yarn and updated it in the "resolutions" config of the package.json. As we received a new package-lock.json file from "npm installing" a deep dependancy of many Gov React modules (govuk-colours) gets effected too so I have explicitly updated this as well.

### This PR basically does 4 things:

1. Updates Jest
2. Fixes security vulnerabilities
3. Updates Gov uk colours
4. Removes Yarn

I couldn't really separate these out into separate PR's as one effects the other...


ian-leggett:
> Looks good. Appreciate this wasn't an easy one!!
> 
> I think you just need to add `yarn.lock` to the `.gitignore` file and then we're good.

Thanks Dave and good spot, just updated.

PR #320, opened 2020-09-10
Accessibility work: Add accessibility attributes to the ActivityFeed's filters

LaurenQurashi:
As part of the accessibility work currently going on, a sample of 25 pages are having accessibility fixes made to them before audit. One of the pages in the sample uses the ActivityFeed filters which are still hosted in this project. 

This PR adds some `for` and `id` attributes, and `aria-label` attributes to the ActivityFeed filters, in order to make them accessible.

### DIFF
--- a/src/activity-feed/ActivityFeedCheckbox.jsx
+++ b/src/activity-feed/ActivityFeedCheckbox.jsx
@@ -42,7 +42,11 @@ const StyledCheckbox = styled(Checkbox)`
 
 const ActivityFeedCheckbox = (props) => {
   const { children, ...rest } = props
-  return <StyledCheckbox {...rest}>{children}</StyledCheckbox>
+  return (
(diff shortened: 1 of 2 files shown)


src/activity-feed/__tests__/ActivityFeedApp.test.jsx
```diff
@@ -130,7 +130,7 @@ describe('ActivityFeedApp', () => {
       )
 
       wrapper.find('select').simulate('change', {
-        target: { value: 'externalActivity' },
+        target: { value: 'externalActivity', id: 'externalActivityId' },
```
paulgain, line 133: If you change `externalActivityId ` to `whateverYouLike` all tests still pass. I think it's safe to remove this `id`.

paulgain (changes requested):

PR #325, opened 2020-11-30
Add deprecation notice

cgsunkel:
We are going to deprecate this project next Monday (7th December). I've added a notice of this to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@
 [codecov-image]: https://codecov.io/gh/uktrade/data-hub-components/branch/master/graph/badge.svg
 [codecov-url]: https://codecov.io/gh/uktrade/data-hub-components
 
+-------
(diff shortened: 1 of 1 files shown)


ambrozic (approved):

tomkingston (approved):
Looks good to me. Presumably you will also deprecate the package on npm to make sure that there is a warning message when people install it like it says here: https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

cgsunkel:
> Looks good to me. Presumably you will also deprecate the package on npm to make sure that there is a warning message when people install it like it says here: https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

I don't have the permissions for the NPM package unfortunately, I will check with Paul (who I believe is the only current dev with access)

PR #326, opened 2020-12-07
Update deprecation notice

cgsunkel:
The deprecation day is here for the component library, and I've updated the deprecation notice accordingly. In case there are any other users (or if someone stumbles across the repo and wants to use it), I've added links to the FE Storybook site and components folder.

Once this PR is merged, I will archive the repo.


ian-leggett (approved):
So nice to see this 👍

paulgain (approved):

